PR #10, opened 2021-03-13
Readme: Add how to get started

scamille:
Edit: I need to test this first :-)

PR #12, opened 2021-03-27
Readme: Add how to get started

scamille:
New version of #10

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,5 +38,5 @@ Read 10 bytes from DataBlock 87 starting at address 54
      StartByte = 54
  };
  await connection.ReadAsync(dataItem);
- Console.WriteLine($"Read data: {dataItem.Value}");
+ Console.WriteLine($"Read data: {BitConverter.ToString(dataItem.Value)}");
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,3 +20,23 @@ A good description is available at [The Siemens S7 Communication - Part 1 Genera
...
+Read 10 bytes from DataBlock 87 starting at address 54
+```
+ var dataItem = new DataBlockDataItem<byte[]>
+ {
+     DbNumber = 87,
+     Length = 10,
+     StartByte = 54
+ };
+ await connection.ReadAsync(dataItem);
+ Console.WriteLine($"Read data: {dataItem.Value}");
```
mycroes, line 41: Doesn't this just print Array`1?

scamille: You are right.

I don't want to add too much boilerplate code here, do you mind if I use [BitConverter.ToString(byte[])](https://docs.microsoft.com/en-us/dotnet/api/system.bitconverter.tostring?view=net-5.0#System_BitConverter_ToString_System_Byte___) in the example here?

mycroes (approved):

mycroes:
Yes this is fine. Seems I messed up the validations, but will merge this as is.

Issue #14, opened 2021-03-29
Expose protocol overhead size for requests


scamille:
I want to split my large data reading/writing requests into smaller chunks, similar to https://github.com/S7NetPlus/s7netplus/blob/051091919fa80b03989f4375e364ebc4b02aabe7/S7.Net/PlcAsynchronous.cs#L106
For that I need to know at what size to split my data into chunks.

I can get at the PDU from the IS7ConnectionParameters, that's great.

What I am missing are those magic numbers of protocol overhead for a request & per data item. I don't want to hardcode these, and since you have such well-organized data types for all your protocol code, would it be possible to expose the size of those structs somewhere, eg. in a static class?

scamille:
Oh and I guess just to be on the safe side: Is the overhead of sending the read request larger than the overhead required for receiving the response?

In the end the only thing that I really care about is the maximum number of bytes payload I can put into a request, whatever part in there is the limiting factor.

PR #16, opened 2021-03-29
Expose protocol struct sizes

mycroes:
Sample fix for #14.

@scamille Let me know what you think (didn't have the time to finish this right now).


scamille:
Sally7.Protocol.SizeOf looks nice. I would just blindly assume that my available payload in bytes for a read request is:

var sizeAvailable = connection.Parameters.MaxPDU - SizeOf.ReadRequest - NumberOfDataItems * SizeOf.DataItem;

Do I need to do something with SizeOf.Tpkt?

mycroes:
The assumption is incorrect unfortunately. There's a two (I think) additional structs that are part of the message and there's the TPKT header, although I doubt that the TPKT is part of the PDU constraint. I'm not sure how I'm going to clarify the different message parts, but I'll make sure there's also some convenient property that'll let you worry about DataItems only.

scamille:
Let me know if I should do part of the work for such a "May Payload" calculator. I mainly wanted some of the internal sizes exposed so I don't have to use some magic numbers.

In the end I think I want:
```
public static int MaximumReadRequestPayload(int maximumPduSize, int numberOfDataItems);
public static int MaximumWriteRequestPayload(int maximumPduSize, int numberOfDataItems);

public static int MaximumReadRequestPayload(this IS7Connection connection, int numberOfDataItems) =>
  MaximumReadRequestPayload(connection.Parameters.MaximumPduSize, numberOfDataItems);
public static int MaximumWriteRequestPayload(this IS7Connection connection, int numberOfDataItems) =>
  MaximumWriteRequestPayload(connection.Parameters.MaximumPduSize, numberOfDataItems);
```

mycroes:
I'm almost done, actually tested the maximums as well because I wasn't sure what the boundaries were. I think I'll have time to round it all up tomorrow.

mycroes:
The changes were kinda done, but I wasn't able to commit them earlier. I didn't implement the extensions per your suggestion, because I fail to see a good usecase for it. I can image you'd want to know the maximum available length in bytes for a single `DataItem`, but I don't see how it matters what the available bytes is for 2, because logically you'd only want to cut the size of one of the `DataItem`s. I'm not yet using this myself either, so I'm not sure what would be a nice solution.

PR #18, opened 2021-04-28
Editable S7 Port

mach-sfortier:
Modification to allow changing the default port 102.

For purpose where more than one S7 server are started on a station with different S7 listening port.

### DIFF
--- a/Sally7/Plc/ConnectionFactory.cs
+++ b/Sally7/Plc/ConnectionFactory.cs
@@ -41,9 +41,9 @@ namespace Sally7.Plc
             }
         }
 
-        public static S7Connection GetConnection(string host, CpuType cpuType, int? rack = null, int? slot = null)
+        public static S7Connection GetConnection(string host, CpuType cpuType, int? rack = null, int? slot = null, int? tcpPort = null)
         {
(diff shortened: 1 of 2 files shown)


mycroes:
Hi @mach-sfortier, can you tell me how this works on the PLC side? I'm not a PLC programmer, but I never heard any of the PLC programmers at our company say anything about port numbers (they actually ask me instead...). I'd be willing to merge this if it makes sense in the scheme of Siemens PLC's, if it's only here because it's theoretically possible (port forwarding, software stations) I'm not sure if I want it to take such a prominent place. Please let me know, either I can merge it as is or find another way that at least allows one to change the port.

mach-sfortier:
Hi @mycroes,

This modification isn't for Siemens PLC's.
It's more to use Sally7 with the simulation of multiple Siemens PLC on a computer with only one network card available.
In this case, the simulated Siemens PLC will be started with different port.

Best regards,
Stéphane

scamille:
I guess you could also always imagine a scenario where the network does something to map the true port to a different one, for whatever reasons.

The S7NetPlus library allows for adjusting the port exactly because of the Snap7 unit tests, which don't run on non windows systems without special privileges on low ports. 

That works fine there, so it should work here as well.

Issue #7, opened 2020-06-22
Q: Comparsion to S7netPlus


MiroslavMikus:
Hey,

we are using the [S7NetPlus](https://github.com/S7NetPlus/s7netplus) library in our project. Since you are one of the contributors to the S7NetPlus library and owner of Sally, there is no better person to ask.

- How do these two libraries compare to each other
- In which scenarios should I decide to use one over another

Thanks in advice
Best regards Miro

mycroes:
Hi @MiroslavMikus,

Sorry for my incredibly late response. I have seen your message when you asked your question, but have forgotten to respond. In short, Sally7 has a very confined interface and currently (out of the box) only supports reading and writing of DataItems. However, Sally7 is very fast with reading and writing and offers strongly typed results. Also, Sally7 only offers async methods for reading and writing (because network communication is async).

S7NetPlus on the other hand has a convoluted API and tries to support many features, which in result means there's more code to maintain and it hasn't been properly maintained.

Both libraries work well. If Sally7 covers what you need I highly recommend Sally7. We're using it in production for a multitude of customers where performance matters. Sally7 is easy to consume (because the public API is small). Sally7 will permit you to have read/write cycles within 10ms, perhaps even less.

Let me know if you want more info.

PR #20, opened 2021-10-06
Removes the in modifier, as it's not needed and harms perf by passing as (readonly) reference

gfoidl:
In C# structs are passed to methods by value, so they are "readonly" by construction. Hence there's no need to add the `in` modifier.

`in` may harm perf, as the arguments can't be passed in CPU-registers, rather they have to be passed via stack-allocated memory, which is way more overhead than simple passing by registers.

`in` is only useful when `readonly struct`s need to be passed around, so that C# compiler doesn't need to create a defensive copy.

This PR removes all unnecessary uses of `in`.
Note: technically this is a breaking change, but as the version of the project (according to SemVer) is below 1.0.0 this shouldn't be a problem.

### DIFF
--- a/Sally7/DataBlockDataItem.cs
+++ b/Sally7/DataBlockDataItem.cs
@@ -28,7 +28,7 @@ namespace Sally7
             }
         }
 
-        private DataBlockDataItem(in ConvertToS7<TValue> toS7Converter, in ConvertFromS7<TValue> fromS7Converter, in int elementSize)
+        private DataBlockDataItem(ConvertToS7<TValue> toS7Converter, ConvertFromS7<TValue> fromS7Converter, int elementSize)
         {
             this.toS7Converter = toS7Converter;
             this.fromS7Converter = fromS7Converter;
@@ -75,11 +75,11 @@ namespace Sally7
         TransportSize IDataItem.TransportSize => transportSize;
         VariableType IDataItem.VariableType => variableType;
 
-        int IDataItem.WriteValue(in Span<byte> output) => toS7Converter.Invoke(value, Length, output);
+        int IDataItem.WriteValue(Span<byte> output) => toS7Converter.Invoke(value, Length, output);
 
-        void IDataItem.ReadValue(in ReadOnlySpan<byte> input) => fromS7Converter.Invoke(ref value, input, Length);
+        void IDataItem.ReadValue(ReadOnlySpan<byte> input) => fromS7Converter.Invoke(ref value, input, Length);
 
-        private void SetLength(in int newLength)
+        private void SetLength(int newLength)
         {
             if (length == newLength) return;
 
--- a/Sally7/IDataItem.cs
+++ b/Sally7/IDataItem.cs
@@ -14,8 +14,8 @@ namespace Sally7
         TransportSize TransportSize { get; }
         VariableType VariableType { get; }
 
-        int WriteValue(in Span<byte> output);
-        void ReadValue(in ReadOnlySpan<byte> input);
+        int WriteValue(Span<byte> output);
+        void ReadValue(ReadOnlySpan<byte> input);
     }
 
     public interface IDataItem<TValue> : IDataItem
--- a/Sally7/Internal/SocketTpktReader.cs
+++ b/Sally7/Internal/SocketTpktReader.cs
@@ -62,7 +62,7 @@ namespace Sally7.Internal
             return receivedLength;
         }
 
-        private static int GetTpktLength(in ReadOnlySpan<byte> span)
+        private static int GetTpktLength(ReadOnlySpan<byte> span)
         {
             try
             {
--- a/Sally7/Protocol/BigEndianShort.cs
+++ b/Sally7/Protocol/BigEndianShort.cs
@@ -8,12 +8,12 @@ namespace Sally7.Protocol
         public byte High;
         public byte Low;
 
-        public static implicit operator BigEndianShort(in int value) =>
+        public static implicit operator BigEndianShort(int value) =>
             new BigEndianShort {High = (byte) (value >> 8), Low = (byte) value};
 
-        public static implicit operator BigEndianShort(in byte value) => new BigEndianShort {High = 0, Low = value};
+        public static implicit operator BigEndianShort(byte value) => new BigEndianShort {High = 0, Low = value};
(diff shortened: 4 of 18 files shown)


mycroes:
Hi Günther,

First of all, thanks for your contribution. I always appreciate new people chiming in and your GitHub profile alone (as short as it is) put a smile on my face 😄 

Now on to the PR: The intention of using `in` was twofold, one for declaring the arguments as invariants and the other to actually improve performance. The former is absolutely not of utmost importance for this library. I consider this a small codebase, so I'm not worried about dropping in with regards to keeping the variables invariants. The latter is based on the fact that arguments can/will be passed by reference, I'm not sure if the performance benefit of that is something I made up or something I've read about. However, your statement about the fact that this causes unnecessary stack allocations at least sounds plausible, could you point me to some more info on that subject? Also, did you try benchmarking this? Would be nice to see what difference this makes as well (even if it's just a contrived sample to see the benefits).

Regards,
Michael

mycroes:
Hi,

> > dropping in with regards to keeping the variables invariants
> 
> As structs are value types, and passed by value they behave like invariants. For Span and ReadOnlySpan the JIT has special handling, which will be prohibited by `in`.

I meant actual invariants in the sense that the compiler will prohibit you from re-assigning the variable if it's declared as `in`, but as I said that's not a major concern.

> > could you point me to some more info on that subject?
> 
> Let's look at
> ...

I was expecting a link to a post, but you're posting a sample that might as well deserve a post in itself 😄. Nice explanation, I guess this says enough about the impact of this change.

> I like this project, it fits my needs very well.

Care to detail that any further? In all honesty I'm aware of a few users only. I'm using this for Viscon machines as part of my day job, I know someone is using it for a trash sorting machine and recently I was approached by designers of a low code platform to use this as alternative to libnodave, so I'd love to hear more success stories.

Besides success stories, I heard of one unconfirmed issue with the latest Sally7 release that it might cause multiple write operations for a single write action. In all honesty at Viscon we're still using the latest version that doesn't include the concurrent operation support, I haven't yet figured how I'm going to test this behavior.

> As I'm always interested how things are done, and thanks to open source, I can look into the code. If I spot some things I like to change or improve, then it's time to try it and create a PR 😉

+1 from me. I must admit I like to keep tight control of projects I maintain, yet I absolutely don't feel I'm the best programmer in/of whatever. I really appreciate contributions like these though and I especially love it when I can actually learn something from it, so thank you very much for that.

mycroes (approved):
Only comment I have on the commit is the message, I try to follow [this post](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) as guideline. Since I don't think I actually have contributor guidelines I'll happily accept this change 😉

scamille:
> Sounds like a race condition. I need to have a closer look into the code, and I can try to reproduce that behavior with SoftPlc. We are talking about the change introduced by [279ddc4](https://github.com/mycroes/Sally7/commit/279ddc4d3983371e19741007d8bce451b5617195)?

Yeah I ran into problems with that when I tried to deploy it during one of our projects, but it might have been something completely different. The problems definitely went away after switching the driver form Sally7 back to the other S7 library we use.
Unfortunately I never had the time to fully investigate what was causing the problems :/ I would have loved to add a runtime behavior switch to optionally guard all Sally7 calls behind a lock (async based) to see if that changed the behavior.

It might also have been the case that our PLC programmers just were not prepared to handle the concurrent requests, and there is no problem with the Sally7 implementation at all. Again, unfortunately we couldn't properly analyze anything of this.

mycroes:
> > Care to detail that any further?
> 
> Without going into too much details (I'm not allowed to talk about this), but it's quite aged machine in the chemical industry with a very old S5-948 and a S7-315 PLC. Both are accessible via ISO over TCP, and the task is to monitor and collect the machine values (data blocks). So at the moment it's read-only, with plans to enter write-mode later too.

Thanks for this glimpse, it's nice to hear the area of use and the technology it's applied to.

> When that project was started, I evaluated some .NET PLC libraries (S7netplus, Sally7, Sharp7) and had a look at proprietary ones too. But I like open source, as mentioned above. The choice was Sally7, mainly to reasons you outlined in [#7 (comment)](https://github.com/mycroes/Sally7/issues/7#issuecomment-742827578) too (so strong typed results, and async + easy to use).
> 
> So far everything the project needs is covered by Sally7.

Basically I see Sally7 as the single best choice when it covers the required feature set. I also don't mind expanding on features as needed, but for my own use I still haven't required more than is in here.

> For local and CI testing https://github.com/fbarresi/SoftPlc is used to "emulate" the PLC.

I still intend to tackle this with Sally7 as well at some point, it doesn't have any priority though.

> > heard of one unconfirmed issue with the latest Sally7 release that it might cause multiple write operations for a single write action
> 
> Sounds like a race condition. I need to have a closer look into the code, and I can try to reproduce that behavior with SoftPlc. We are talking about the change introduced by [279ddc4](https://github.com/mycroes/Sally7/commit/279ddc4d3983371e19741007d8bce451b5617195)?

Exactly. I'm used to multithreading and synchronization primitives and all that, doesn't mean I don't make mistakes either. Given that I also write most of my open source code late in the evening I'd say there's a fair chance of bugs popping up, there's room for improvement there as well.

> > so thank you very much for that.
> 
> That's why I ❤️ open source. With the knowledge of many one can learn from each other, and improve a product. But there's no need to thank me, as it's quite opposite. Sally7 is built on your knownledge and I'm happy to be able to use it.

Well again, thanks. My knowledge has been made possible by others sharing their knowledge and now you're a part of that as well.

gfoidl:
> Sounds like a race condition. I need to have a closer look into the code, and I can try to reproduce that behavior with SoftPlc. We are talking about the change introduced by 279ddc4?

When looking at the code, and stepping with the debugger through it I can't spot a place where a race could occur (assuming the types used from .NET are correct). Also tried to repro this behavior, but without any success.
I moved some things around (967f9504f6c6fa8e1ad2f5a6f486285664d135b6 built on top of https://github.com/mycroes/Sally7/pull/21), also can't repro it.
BTW: should this commit be brought into _main_? (I like that the Channel<T>s are hidden as implementation detail -- my motivation was to try a SemaphoreSlim, but this allocates more because of Task vs. ValueTask).

PR #21, opened 2021-10-06
Some low hanging fruits optimizaitons

gfoidl:
From the project description: "with a focus on performance", so let's squeeze out a bit more...

---

### Throw-helper pattern

This pattern uses local functions and goes like this:
```c#
if (condition)
{
    Throw();
    static void Throw() => throw new Exception(...);
}
```
The JIT won't inline methods that will never return (so methods that `throw`), and with that pattern the quite costly setup of the exception-object is moved to a differnt method.

The effect can be best seen at looking to the machine code.
[normal exception pattern](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNXjxldskhrgxQOYDAxH583Zdj4Qu8jBzE0A3jQZemAZiYBWey4RABMYAKDQmA1qb18GYABPDHC2VUUYEJi4zzjiP1hsEIdJRKYUBgBBXFwZDAAKYkDeByiGADNuMxbgiDCASly4rw9Y4fHedoZG5ta+8IBCAF4Orsj5/oYMAAsoCAB3Bi4YQ4BRBBUxIPqAEgAic7EYM0zVrm6HBkgwhjdO9/WYQAvmgGLAVLwFK83E1ATAgSw7v1suNhpNpmlalAoSEGMsGHRNjs9odjmcLjArj1bg8EE8XrjYFicQTQeCYJDoZiMiEgUiUeMgTQhdQtDooHoDDAuBx8BEem0QAlkjAhr81XE0kUGCs6Ag6OgNd4AOpQXgpHUE/X+NBGryWaxcWxBJwuS169p0NVAoA===) or [another example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjQDaUjFA5gMAGWwBPCBwwAKQybMBpXlwAmLGQEcOMHr2wAbNAwAK2GAA1gwAvAzkAJQAujT2GGJcPgy42roYDAAqEsEYNADeNAwlDBJQvLLYiQyQXGkMwEY1QrgQAGpiuLwQXOEMAMyq1MWlxAONzTAMnVDdvcOlTBNNNWxKcjAOi2MTAEK8AOYAoo5eXFIAFtCZ+h6HGJfDoyXjDLDYDr3eRkwoDABBXC4MQWKIvBhFahLJa8ABmDHMs3mfQAhBFWh0uj0uFEGI9xAB3BhcGDE44IRQSPi9cwAIikEhgYAY8gg3mqOJA2X8NiyDEuME+YgFeAY9kgUFgelZ2N6LDpUR2MPFCPM6xBUFkWwY6IYdDxBIgxNJ5MpMGpOPpjOZrJ63ggnN63KyvP5sE12oc4twJIgmToCqVENharuXAelxYAAkjpdwhFGAAyJMMcORlj6Y0MAA8DAA7IbLkSSWSGBSqTSuNamSy2Ryqy63QxvPdHj7UvgfK2oPjLtg+vmg8qGABfGjjkbUTTpPRWUwWee2exOVzuTw+PyBEL9aJxWgTNI6GUHE5nAdXG6FCFvVbTWOHJ43lZTNPG57Q3ZMchIcX4CTeAovCZJIYjVNADCnqcDjnJeUAWAkrI+O4eJhAAfCGpSmpBRzQbB1zwZCD7xhE5h3ni5hVN47gMGhaEMAAHFEfhZsSpHkUh1EwKOH5LG8zC/vwAFASBTJQOBvZQeeFwERYd6cSh4T0dhUkwReslEXG/R0Cx2YRFR7g8Won6vBMAl/sJYDAQwoHiRgEEJOYqn4TcJK4Khym4DGWk5nmDEMIQbmZu+xl8RMEDalKvAONMzCMFkEBaBUEbmB5iLmAkRa8O5LAJUl9iHKlwyjkAA)
[with throw-helper](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNXjxldskhrgxQOYDAxH583Zdj4Qu8jBzE0A3jQZemAZiYBWey4RABMYAKDQmA1qb18GYABPDHC2VUUYEJi4z29iP1hsEIdJRKYUBgBBXFwZDAAKYkDeByiGADNuMxbgiDCASlyvD1i47152hkbm1r7wgEIAXg6uyLnB0bGGEa2tgBUACygIAHdptbC0Fa5u2YHs3bjmJHKGQ+OzpovwmAQxGDMmSuXx6bWwZg4+n6DEWAD4GBgjqcGFwYCcGABRBAqMRBeoAEgARFj/oCQtdblwGJAwttfqSUiEAL5XWAqXgKTLbcHOfRMliE/oPLZMzSbMYTKZpWpQTnkpYMOgbR47R7ed6nerSjIhIVDNXPV4as5JFIMWAyuXQuEIpHo1HorE4vFEkkAxnm9Kyrl0VkAmAcrluC06pmC4VjUXiqNRrQ6KB6AwwLgcfARUFzBggBLJGD61VbNJFGGKhB0dD6uIAdSgvDNyzoZf8aEr3ks1i4tiCThcJcb7To+qZQA=) or [another example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjQDaUjFA5gMAGWwBPCBwwAKQybMBpXlwAmLGQEcOMHr2wAbNAwAK2GAA1gwAvAzkAJQAujT2GGJcPgy42roYDAAqEsEYNADeNAwlDBJQvLLYiQyQXGkMwEY1QrgQAGpiuLwQXOEMAMyq1MWlxAONzTAMnVDdvcOlTBNNNWxKcjAOi2MTAEK8AOYAoo5eXFIAFtCZ+h6HGJfDoyXjDLDYDr3eRkwoDABBXC4MQWKIvBhFahLJa8ABmDHMs3mfQAhBFWh0uj0uODoTCSlCCQSspdxAB3cxRHbEsbkJB/bJkiCUqLhAB8DEeFIYXBg5IYxwQigkfF65gARFIJDAwAx5BBvNUcSBsv4bFkGJcYJ8xFq8Ax7JAoLA9PLsb0WBLqRClgBfNT4gnwxHrEFQWRbBjohh0PG0om00qkilUmlB5gM4j/EMsqkcrnMgV8gVCkViriS6Wy+U9bwQZW9VVZdWa2Duz0OQ24XkQTJ0K02p0wh0jZuwhGWe6PFgACSOl3CEUYADIRww7lwHpcWPoWQwADwMADs/uJgaDTNDTc3r3pjNjrIT3PnKcFwpgopxWZlcoVSozxdLDG83cHvBrBB8r6giewfWXRtw3tCFW1bDQtB0PQrFMCwYNsewnFcdxPB8PxAhCfpojiWgJjSKDMgOE4zn/K4bkKCE3lWaZ+0OJ5KJWKYJxZZ5mzeSNDXwCRvAUXhMkkMRqmgBgiNOBxzjIqALASeUfHcNkwnZW1SjPUSSIua4pMhWjBwicxqLZcwqm8dwGHZTkAA4oj8OcBT0gzZJMmA7VYpZ2P3fhuN4/iZSgITfzU8TSM0ixqMc+SE1Uo4xIkkLtIHfo6Bs+cImM9wXMdNyJg4zyeLAPiGAEvyMGEhJzEC2Kbl5XAFM5eo+wShclwshhCGq2cWMy3ZCs9E1eAcaZmEYLIIEg+xDnjRTEXMBI2UeD8ohYEaxqnMMaDtIA==)

So without that pattern, these are huge methods that won't be inlined.
By applying the throw-helper pattern, the methods become small so that inlining makes sense. Due the `if`s I'm not entirely sure if the JIT will actually inline these methods, so I gave the JIT a hint by appyling MethodImpl-options.

### BinaryPrimitives

Just look at the [machine code](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5il3YBLAHbYoATwDcNGsQDMTUgwDCDAN40GWhgAcoggG7YMMJuSRMUKiMIN8MAFQgAZGwHMAFLC4MANu4YjXw4YNAYAJRhsABMAeWFfcQBlHWxhAB5gcRMAPgYRHQ4MMJEMPxhhNwwACwBKTW0Nam0WwOxg0wBeBg9/StqejlKBjwKigG06AF0GdPSGUitCfOFCjHHyGbmGcwZlsfXSLfmADj2VtfG5KYHtuSVCBtbWjyHhDBGD8ZRjhaWLiYAVl+u32qwmSF+ZzBlwA7DdpM1tABfJ4MNF6QzGUzMCzEKzKGx2KCOFzucheGA+PpuNodErCCJROIJZKpDJZXIA4orFZlXwVKp1NEtTp5IIhBjdNgiMTiAAK+nwggwhhgAkiMQAku9UDK3ABRYTRQRpUbgj6I5FAA) to see the difference.

The JIT is able to emit special instructions (in the example for `long` a `bswap`-x86 instruction) which results in way less code that needs to be executed.

### typeof(TValue)

The JIT is able to dead code eliminate branches, but only when `typeof(TValue)` is used directly. See [this example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5il3YBLAHbYoATwDcNek1YAlDsIyD8MFgGEI+AA6CANnwDKfAG6CwMXNNqMjACzE6AMtmAtFy1TBsyAzEykDBoMAN40DJEMANoAUoIYAOIwwnwWABQY4jowEFzpIhgAlEUAuhFRxAGFDMkYRoIAXjAAPAAqAHzpRRWR4dRRgwymYgxZOQwAvGPZufltRTZDUYJcDJmzLACSuABq2PocMG2zPQPLYb0XkcQA7AwAqsK42LwsDc0A8lztXYtXgwAvgCojB9LgYAxVutxpDJtNYXl0sw6CUQX10UM7gwUKQlhdgedlpiMPYoBAAO4MVJUgByEHqHB0OmgGBgABMAKIISw6FQQYTdfGRQmDK5xBLJVJQDKI/KFErlIlMarKWowepNGDkX7dK79a7QjY5JELbZ7A5HE45M7XA3XSr3J4vN4fGDfXX/ZVAzFgiFQtbGubpBZTBGzJEotHeqL2h1Me644VDUUXElkynUmB0hlGJksqBsrk8mB8wQCoVXQmAoA===) where the whole code is eliminated and the (expected) constant value is used directly.

---

That and https://github.com/mycroes/Sally7/pull/20 was the eye-balling part of optimizations...further ones are harder to play (and the code-base looks quite good 👍🏻).


Sally7/S7Connection.cs
```diff
@@ -100,8 +101,7 @@ public async Task OpenAsync()
             await TcpClient.ConnectAsync(host, IsoOverTcpPort).ConfigureAwait(false);
             var stream = TcpClient.GetStream();
 
-            // TODO: use memory from the pool
-            var buffer = new byte[100];
+            var buffer = ArrayPool<byte>.Shared.Rent(100);
```
gfoidl, line 104: This will actually rent an array of size rounded up to the next power of two, so the length will be 128.
This shouldn't be problem, as the implementation in the S7Connection is defensive for that.

Sally7/S7ConnectionHelpers.cs
```diff
@@ -59,8 +59,8 @@ public static int BuildReadRequest(in Span<byte> buffer, in IReadOnlyList<IDataI
         public static int BuildWriteRequest(in Span<byte> buffer, in IReadOnlyList<IDataItem> dataItems)
         {
             var span = buffer.Slice(17); // Skip header
-            span[0] = (byte) FunctionCode.Write;
             span[1] = (byte) dataItems.Count;
+            span[0] = (byte) FunctionCode.Write;
```
gfoidl, line 63: This is just a mini-optimization. By flipping the order (the largest index first), the JIT only needs to emit one bound check instead of two.

Sally7/ValueConversion/FromS7Conversions.cs
```diff
@@ -113,13 +93,15 @@ private static void ConvertToByteArray<TTarget>(ref byte[]? value, in ReadOnlySp
...
         private static void ConvertToBoolArray(ref bool[]? value, in ReadOnlySpan<byte> input, in int length)
-        {
-            value = new BitArray(input.Slice(0, (length + 7) / 8).ToArray()).Cast<bool>().Take(length).ToArray();
-        }
+            => value = new BitArray(input.Slice(0, (length + 7) / 8).ToArray()).Cast<bool>().Take(length).ToArray();
 
         private static void ConvertToString(ref string? value, in ReadOnlySpan<byte> input, in int length)
         {
+#if NETSTANDARD2_1_OR_GREATER
+            value = Encoding.ASCII.GetString(input.Slice(2, input[1]));
```
gfoidl, line 101: We can avoid the allocation of the temporary array on NS 2.1 or newer (there won't be any newer anyway for .NET Standard as .NET 5 onwards is the future).

Sally7/ValueConversion/ToS7Conversions.cs
```diff
@@ -153,6 +137,22 @@ private static int ConvertFromString(in string? value, in int length, in Span<by
                 return 2;
             }
 
+#if NETSTANDARD2_1_OR_GREATER
+            var maxByteCount = Encoding.ASCII.GetMaxByteCount(value.Length);
+            Span<byte> span = maxByteCount <= 256
```
gfoidl, line 142: I'm not sure if this is overkill here...
Is string-handling in PLC that common? I don't know.

Anyway...if the string is short enough we stack-allocate, otherwise just do a regular allocation. One could rent a buffer from the array-pool, but this makes the code a bit more unreadable though.

mycroes:
Really like these changes and the samples, will do another review round on latest update ASAP.

mycroes:
> From the project description: "with a focus on performance", so let's squeeze out a bit more...
> 
> ### Throw-helper pattern
> This pattern uses local functions and goes like this:
> 
> ```cs
> if (condition)
> {
>     Throw();
>     static void Throw() => throw new Exception(...);
> }
> ```
> 
> The JIT won't inline methods that will never return (so methods that `throw`), and with that pattern the quite costly setup of the exception-object is moved to a differnt method.
> 
> The effect can be best seen at looking to the machine code. [normal exception pattern](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNXjxldskhrgxQOYDAxH583Zdj4Qu8jBzE0A3jQZemAZiYBWey4RABMYAKDQmA1qb18GYABPDHC2VUUYEJi4zzjiP1hsEIdJRKYUBgBBXFwZDAAKYkDeByiGADNuMxbgiDCASly4rw9Y4fHedoZG5ta+8IBCAF4Orsj5/oYMAAsoCAB3Bi4YQ4BRBBUxIPqAEgAic7EYM0zVrm6HBkgwhjdO9/WYQAvmgGLAVLwFK83E1ATAgSw7v1suNhpNpmlalAoSEGMsGHRNjs9odjmcLjArj1bg8EE8XrjYFicQTQeCYJDoZiMiEgUiUeMgTQhdQtDooHoDDAuBx8BEem0QAlkjAhr81XE0kUGCs6Ag6OgNd4AOpQXgpHUE/X+NBGryWaxcWxBJwuS169p0NVAoA===) or [another example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjQDaUjFA5gMAGWwBPCBwwAKQybMBpXlwAmLGQEcOMHr2wAbNAwAK2GAA1gwAvAzkAJQAujT2GGJcPgy42roYDAAqEsEYNADeNAwlDBJQvLLYiQyQXGkMwEY1QrgQAGpiuLwQXOEMAMyq1MWlxAONzTAMnVDdvcOlTBNNNWxKcjAOi2MTAEK8AOYAoo5eXFIAFtCZ+h6HGJfDoyXjDLDYDr3eRkwoDABBXC4MQWKIvBhFahLJa8ABmDHMs3mfQAhBFWh0uj0uFEGI9xAB3BhcGDE44IRQSPi9cwAIikEhgYAY8gg3mqOJA2X8NiyDEuME+YgFeAY9kgUFgelZ2N6LDpUR2MPFCPM6xBUFkWwY6IYdDxBIgxNJ5MpMGpOPpjOZrJ63ggnN63KyvP5sE12oc4twJIgmToCqVENharuXAelxYAAkjpdwhFGAAyJMMcORlj6Y0MAA8DAA7IbLkSSWSGBSqTSuNamSy2Ryqy63QxvPdHj7UvgfK2oPjLtg+vmg8qGABfGjjkbUTTpPRWUwWee2exOVzuTw+PyBEL9aJxWgTNI6GUHE5nAdXG6FCFvVbTWOHJ43lZTNPG57Q3ZMchIcX4CTeAovCZJIYjVNADCnqcDjnJeUAWAkrI+O4eJhAAfCGpSmpBRzQbB1zwZCD7xhE5h3ni5hVN47gMGhaEMAAHFEfhZsSpHkUh1EwKOH5LG8zC/vwAFASBTJQOBvZQeeFwERYd6cSh4T0dhUkwReslEXG/R0Cx2YRFR7g8Won6vBMAl/sJYDAQwoHiRgEEJOYqn4TcJK4Khym4DGWk5nmDEMIQbmZu+xl8RMEDalKvAONMzCMFkEBaBUEbmB5iLmAkRa8O5LAJUl9iHKlwyjkAA) [with throw-helper](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNXjxldskhrgxQOYDAxH583Zdj4Qu8jBzE0A3jQZemAZiYBWey4RABMYAKDQmA1qb18GYABPDHC2VUUYEJi4z29iP1hsEIdJRKYUBgBBXFwZDAAKYkDeByiGADNuMxbgiDCASlyvD1i47152hkbm1r7wgEIAXg6uyLnB0bGGEa2tgBUACygIAHdptbC0Fa5u2YHs3bjmJHKGQ+OzpovwmAQxGDMmSuXx6bWwZg4+n6DEWAD4GBgjqcGFwYCcGABRBAqMRBeoAEgARFj/oCQtdblwGJAwttfqSUiEAL5XWAqXgKTLbcHOfRMliE/oPLZMzSbMYTKZpWpQTnkpYMOgbR47R7ed6nerSjIhIVDNXPV4as5JFIMWAyuXQuEIpHo1HorE4vFEkkAxnm9Kyrl0VkAmAcrluC06pmC4VjUXiqNRrQ6KB6AwwLgcfARUFzBggBLJGD61VbNJFGGKhB0dD6uIAdSgvDNyzoZf8aEr3ks1i4tiCThcJcb7To+qZQA=) or [another example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAJI8YUCAAcAymIBuvMDFwBuGjQDaUjFA5gMAGWwBPCBwwAKQybMBpXlwAmLGQEcOMHr2wAbNAwAK2GAA1gwAvAzkAJQAujT2GGJcPgy42roYDAAqEsEYNADeNAwlDBJQvLLYiQyQXGkMwEY1QrgQAGpiuLwQXOEMAMyq1MWlxAONzTAMnVDdvcOlTBNNNWxKcjAOi2MTAEK8AOYAoo5eXFIAFtCZ+h6HGJfDoyXjDLDYDr3eRkwoDABBXC4MQWKIvBhFahLJa8ABmDHMs3mfQAhBFWh0uj0uODoTCSlCCQSspdxAB3cxRHbEsbkJB/bJkiCUqLhAB8DEeFIYXBg5IYxwQigkfF65gARFIJDAwAx5BBvNUcSBsv4bFkGJcYJ8xFq8Ax7JAoLA9PLsb0WBLqRClgBfNT4gnwxHrEFQWRbBjohh0PG0om00qkilUmlB5gM4j/EMsqkcrnMgV8gVCkViriS6Wy+U9bwQZW9VVZdWa2Duz0OQ24XkQTJ0K02p0wh0jZuwhGWe6PFgACSOl3CEUYADIRww7lwHpcWPoWQwADwMADs/uJgaDTNDTc3r3pjNjrIT3PnKcFwpgopxWZlcoVSozxdLDG83cHvBrBB8r6giewfWXRtw3tCFW1bDQtB0PQrFMCwYNsewnFcdxPB8PxAhCfpojiWgJjSKDMgOE4zn/K4bkKCE3lWaZ+0OJ5KJWKYJxZZ5mzeSNDXwCRvAUXhMkkMRqmgBgiNOBxzjIqALASeUfHcNkwnZW1SjPUSSIua4pMhWjBwicxqLZcwqm8dwGHZTkAA4oj8OcBT0gzZJMmA7VYpZ2P3fhuN4/iZSgITfzU8TSM0ixqMc+SE1Uo4xIkkLtIHfo6Bs+cImM9wXMdNyJg4zyeLAPiGAEvyMGEhJzEC2Kbl5XAFM5eo+wShclwshhCGq2cWMy3ZCs9E1eAcaZmEYLIIEg+xDnjRTEXMBI2UeD8ohYEaxqnMMaDtIA==)
> 
> So without that pattern, these are huge methods that won't be inlined. By applying the throw-helper pattern, the methods become small so that inlining makes sense. Due the `if`s I'm not entirely sure if the JIT will actually inline these methods, so I gave the JIT a hint by appyling MethodImpl-options.

One thing I was also wondering about (based on these changes and recent experience) is moving the throws to a separate class entirely. One of the main benefits is that all error messages end up in a  single class, so it's easier to maintain standards across messages. It also has the added advantage of having a description in the form of the method name. I'd assume this has the same effects as the local static methods. I remember having seen this before in other projects (I believe EasyNetQ has a `ThrowHelpers` class as well), a quick google lead to this [stackoverflow question](https://stackoverflow.com/questions/1980044/when-should-i-use-a-throwhelper-method-instead-of-throwing-directly) with some nice responses about such an approach. I also really like the accepted answer from Mark Seemann (@ploeh), but that doesn't yield the same optimizations as having the throw externally. Given the performance focus I'm totally fine with this tradeoff.

Also, when actually naming the methods consistently we don't have to worry about i.e. `ThrowInvalidTpktLengthReceived` showing up at the top of the stack trace, it'll actually be really informative.

I don't see any reason to include above into this commit, but I would appreciate if you guys can share your thoughts on this subject.

gfoidl:
> moving the throws to a separate class entirely

In short: I'm fine with that suggestion, and will push a commit for this soon.

> I'd assume this has the same effects as the local static methods.

Yeah, has the same effect.

---

In my project I follow this rough guideline for the throw-helpers (on a per "exception-type" basis):
* if it's used only once in a method -> local function
* if it's used several times in a class -> private static method
* if it's used in several classes -> separate ThrowHelper-class

> is that all error messages end up in a single class, so it's easier to maintain standards across messages

This is actually a good point. Also in regards of potential localization / globalization tasks.
A counter point is that the code doesn't have great locality anymore. Take e.g.
https://github.com/mycroes/Sally7/blob/263de5c70ce8ddee1f21a68ea576a9f7ea931533/Sally7/Protocol/IsoOverTcp/Tpkt.cs#L19-L35
where all code is within the same method and file. With a separate class one has to navigate to two files.

But (now) I think the benefits of having a separate class is preferable (here).

gfoidl:
> but I think I'll be able to merge and release on Wednesday at the latest.

I'm out of office until Wednesday, soll will address your points then. This means also you have enough time with the release 😃 
(didn't read other points now, will do on Wed)

gfoidl:
Thanks. Very valuable conversation -- in essence: I like the collobaration here ❤️

PR #30, opened 2023-05-04
Cancellation

mycroes:


mycroes:
@gfoidl Would you care to review this?

Sally7/S7Connection.cs
```diff
@@ -104,58 +100,98 @@ public void Dispose()
...
+                    Parameters = new S7ConnectionParameters(pduSize, maxRequests);
+                    bufferSize = Parameters.GetRequiredBufferSize();
+
+                    memoryPool ??= new Sally7MemoryPool(bufferSize);
+
+                    requestExecutor = executorFactory.Invoke(this);
+                }
+                finally
+                {
+                    ArrayPool<byte>.Shared.Return(buffer);
```
scamille, line 150: Too bad there isn't a IDisposable helper to take care of this...

I really don't miss the days where it is so hard to write a simple destructor allowing for powerful RAII classes, or quickly use a ScopeGuard with a lambda as in C++ :-)
(But of course other things like async and LINQ are sorely missed)

Sally7/S7Connection.cs
```diff
@@ -167,11 +203,26 @@ public async Task WriteAsync(IDataItem[] dataItems, CancellationToken cancellati
...
+            try
+            {
+                // The response will only be written after the request has been sent. At that point we no longer
+                // care about the request contents, so we use a single buffer only.
+                Memory<byte> response = await executor.PerformRequest(mem.Slice(0, length), mem, linkedToken)
+                    .ConfigureAwait(false);
+
+                S7ConnectionHelpers.ParseWriteResponse(response.Span, dataItems);
+            }
+            catch (OperationCanceledException) when (linkedToken.IsCancellationRequested)
```
scamille, line 218: Can you explain why it is necessary in the first place to re-throw with the appropriate token?

I would expect in most cases if a OperationCancelledException is thrown by the downstream code having called `linkedToken.ThrowIfCancellationRequested();` that this would already get the exception for whichever of the 2 linked tokens was cancelled.

Or is this solely about ordering the user cancellationToken's cancellation state as higher than the Timeout Exception?

As far as I remember, cancellation in .NET was a best effort thing anyway, and so I am wondering why I as a caller would be able to rely on getting a cancellation exception from my token at all costs - instead of just being able live with it, no matter whether the cancellation came from my own token or from the configured timeout.

Sally7/S7Connection.cs
mycroes: I'm not sure why they chose this pattern, but creating a disposable pattern for this would be easy-peasy of course. Perhaps the disposable pattern has more of an performance impact, or is often not applicable because the buffer lives longer than the method creating it.

Sally7/S7Connection.cs
mycroes: You'd actually get an exception with `linkedToken` as `CancellationToken`, which makes it impossible to detect if the cancellation happened due to the token being passed in begin cancelled. I guess for day to day usage it's perhaps irrelevant where the cancellation came from, probably especially in the context of Sally7 where it's really shallow. However, when things go south and you start checking the `CancellationToken` of the exception it starts to matter a lot more.

I think I made good effort on the usability by throwing `TimeoutException` when applicable, but to me it appears this is the right way to do it, even if that takes another 2 lines of code.

mycroes:
@scamille Thanks for the initial review!

Sally7/S7Connection.cs
```diff
@@ -69,22 +75,12 @@ public S7Connection(string host, Tsap sourceTsap, Tsap destinationTsap, MemoryPo
...
-        /// </summary>
-        public int SendTimeout
-        {
-            get => TcpClient.SendTimeout;
-            set => TcpClient.SendTimeout = value;
-        }
+        /// <remarks>
+        /// The default value is <see cref="DefaultRequestTimeout"/>.
+        /// </remarks>
+        public int RequestTimeout { get; set; } = DefaultRequestTimeout;
```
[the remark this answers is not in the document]
scamille: Right, but the below -1ms restriction applies to the int version just as well.

My experience with writing code is that using strong types when units are involved is nearly always better, in particular for time durations. I'd much rather use a dedicated duration type rather than a primitive type without a unit, even if it doesn't perfectly match some other API.

You could document the limitations of the `CancelAfter` or just let the `ArgumentOutOfRangeException` take care of that and don't bother - which is good enough in my book.

scamille:
My pleasure. I am quite out of the loop with .NET and also never used the new memory facilities much, so I can't really cover everything. I mainly skimmed over things, and used this to ask some interesting questions :-)

But in general the PR looks great. From a usability perspective, having both user cancellation and proper working timeouts is definitely a very very useful thing to have in this async network code, in particular if you want to scale things properly.

Sally7/S7Connection.cs
mycroes: > As far as I remember, cancellation in .NET was a best effort thing anyway, and so I am wondering why I as a caller would be able to rely on getting a cancellation exception from my token at all costs - instead of just being able live with it, no matter whether the cancellation came from my own token or from the configured timeout.

> I still don't think this will hint you the correct token that caused the cancellation all the time. If both cancellation tokens have been cancelled at the time you are in the catch block, you can't technically tell which one came first. Your code just "arbitrarily" (but that might be a bit of strong phrasing here) prefers to throw from the users cancellation token first if it is canceled.

Yes, cancellation is a best effort thing. because often you want to cancel things you don't control, so instead you cancel further processing of what would happen next. You're right about the race condition with regards to user cancellation and timeout, especially if the passed in token would have an identical timeout as the `RequestTimeout`. Assuming both tokens cancel at the same time there's no way to know which one was first, but I can't imagine why you would still care at that point. And if you would care, it would be fixable by changing the timeout. I have considered not adding the `RequestTimeout` and deferring the responsibility to consumers of the library, but I think this makes it easier for consumers while not polluting the library too much. After all I went from 2 timeout properties to 1 and it actually does something this time around...

mycroes:
> My pleasure. I am quite out of the loop with .NET and also never used the new memory facilities much, so I can't really cover everything. I mainly skimmed over things, and used this to ask some interesting questions :-)

Well this was my first attempt at using memory as well, I think Span actually was one of the reasons for me to start the library. However I just learned I'm not that knowledgeable either. As [arraypool-vs-memorypool-minimizing-allocations-ais-dotnet](https://endjin.com/blog/2020/09/arraypool-vs-memorypool-minimizing-allocations-ais-dotnet) points out there's a difference between `ArrayPool` and `MemoryPool`, the latter of which actually has the dispose pattern you were wondering about. In short, I'd probably be better off using `ArrayPool` instead of `MemoryPool`, so I'll get on that as well.

> But in general the PR looks great. From a usability perspective, having both user cancellation and proper working timeouts is definitely a very very useful thing to have in this async network code, in particular if you want to scale things up (from a number of connections/devices/queries perspective).

The reason for this PR is that it was utterly broken. The cancellation was fine up to a certain point, but not every target framework could make proper use of it. As a result, I had a read action that never finished. Adding the request timeout will handle part of that problem, but it still doesn't cancel a socket read on net461 for instance. Adding the `Close` calls in the right places solves this though, as does running this on net5 or newer. We do have a lot of customer projects running on .NET Framework though, having an updated Sally7 that fixes this problem is the easiest way to handle this (versus updating the target frameworks, which won't always be a realistic option).

Sally7/S7Connection.cs
mycroes: Thanks to your reply I also figured that I should probably close the connection if `Send` is cancelled. We reuse job ID's so if you cancel send, it ends up in the PLC and it replies, then the next job might have taken that ID. Will get that handled as well...

Sally7/S7Connection.cs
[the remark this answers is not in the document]
mycroes: > My experience with writing code is that using strong types when units are involved is nearly always better, in particular for time durations. I'd much rather use a dedicated duration type rather than a primitive type without a unit, even if it doesn't perfectly match some other API.

You're absolutely right. Let's not dwell on why Microsoft chose to use `int` originally, the right thing to do is `TimeSpan`.

> You could document the limitations of the CancelAfter or just let the ArgumentOutOfRangeException take care of that and don't bother - which is good enough in my book.

I'll document and validate, because the exception from `CancelAfter` will only show when sending a request, not when setting the timeout.

Sally7/S7Connection.cs
gfoidl: > often not applicable because the buffer lives longer than the method creating it

Yep, that's the reason -- especially together with the default implementation of the MemoryPool which is / was a motivation to have this type.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
```diff
@@ -92,6 +92,14 @@ public async ValueTask<Memory<byte>> PerformRequest(ReadOnlyMemory<byte> request
                     _ = await sendSignal.WaitAsync(cancellationToken).ConfigureAwait(false);
                     try
                     {
+                        // If we bail while sending the PLC might still respond to the data that was sent.
+                        // This both breaks the send-one-receive-one flow as well as it might end up
+                        // completing a new job that reused the ID.
+#if NET5_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+                        await
+#endif
+                        using var closeOnCancel = cancellationToken.MaybeUnsafeRegister(socket.Close);
```
gfoidl, line 101: As we use `try-finally` already I'd avoid the `using`-pattern here, and create that registration before the `try` and dispose it in the `finally`. Just to avoid the nested try-finally that is generated by the compiler.

gfoidl, line 101: Hm, I'm not sure closing the socket is the correct approach.
In the next statement there's the send-operation on the socket. thus this could throw an `ObjectDisposedException` which is not what one would expect.

At the moment I don't have any better idea (just being back from a few days of holiday my brain isn't up on full power and I need to think the code of Sally7 a bit more...)

mycroes:
> For the `IDisposable`-allocation with the MemoryPool we mitigated this with the `Sally7MemoryPool` (introduced in #23).

You're right. It does mean we can't track use-after-free though (which you can do with allocation per lease), which can happen if the S7 server misbehaves. I'm not really worried about that, but perhaps it would still be a good idea to fix that at some point. However I'd rather fix that by adding some additional checks on the `Request` type instead.

Sally7/Internal/CancellationTokenExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace Sally7.Internal
+{
+    internal static class CancellationTokenExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static CancellationTokenRegistration MaybeUnsafeRegister(this CancellationToken cancellationToken,
+            Action callback)
```
gfoidl, line 11: The signature should be
```c#
public static CancellationTokenRegistration MaybeUnsafeRegister(this CancellationToken cancellationToken, Action<object?> callback, object? state)
```
in order to avoid several allocations -- here and where it's used.
* the closure (allocation of the "display class")
* the allocation for the repeated creation of the delegate

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
gfoidl, line 101: With the suggestion that this extension takes the state as argument too, this should be written as
```suggestion
                        using var closeOnCancel = cancellationToken.MaybeUnsafeRegister(static state =>
                        {
                            Socket socket = state as Socket;
                            socket.Close();
                        }, socket);
```
Thus (amortized) no allocaiton occurs.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
```diff
@@ -121,16 +129,35 @@ public async ValueTask<Memory<byte>> PerformRequest(ReadOnlyMemory<byte> request
...
-                        length = await reader.ReadAsync(mo.Memory, cancellationToken).ConfigureAwait(false);
-                    }
-                    finally
+                    // If we bail while reading we break the send-one-receive-one flow, so we might as well close right away.
+                    // There is minimal risk of closing connections while data was actually received but handling here
+                    // avoids registering on the cancellationToken on every socket call.
+#if NET5_0_OR_GREATER || NETSTANDARD2_1_OR_GREATER
+                    await
+#endif
+                    using (cancellationToken.MaybeUnsafeRegister(socket.Close))
```
gfoidl, line 138: The same as said above applies here.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
```diff
@@ -121,16 +129,35 @@ public async ValueTask<Memory<byte>> PerformRequest(ReadOnlyMemory<byte> request
...
+                        {
+                            try
+                            {
+                                length = await reader.ReadAsync(mo.Memory, cancellationToken).ConfigureAwait(false);
+                            }
+                            catch (Exception) when (cancellationToken.IsCancellationRequested)
+                            {
+                                cancellationToken.ThrowIfCancellationRequested();
+
+                                // Should never happen, but required to satisfy compiler analysis of length initialization.
```
gfoidl, line 151: I'd initialized the `length` in L128 with `0`, so that this statement can be removed.
* feels more natural (for me) than having an artificial exception
* reduced code size
* makes job for JIT easier

gfoidl (commented):
I need to think a bit more about that approach (see comments why I think it's not the ideal solution).

Anyway, your analysis of the latent bug is correct.

Sally7/Internal/NetworkStreamExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+using System.Net.Sockets;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Sally7.Internal
+{
+    internal static class NetworkStreamExtensions
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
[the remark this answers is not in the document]
mycroes: I took inspiration from [RabbitMQ .NET Client](https://github.com/rabbitmq/rabbitmq-dotnet-client), which uses this in quite a lot. I'm not sure what to do with your comment though (I actually expected you might comment on it), would your suggestion be to remove it an let the compiler figure it out, or would you say it doesn't hurt either?

Sally7/Exceptions.cs
```diff
@@ -19,6 +19,12 @@ public class Sally7Exception : Exception
           System.Runtime.Serialization.SerializationInfo info,
           System.Runtime.Serialization.StreamingContext context) : base(info, context) { }
 
+        internal static void ThrowTimeoutIsInvalid(TimeSpan value) =>
```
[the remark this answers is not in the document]
mycroes: I prefer `Assert...` over the `...Invalid` name, but good point about simplification and reuse.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: I always doubt whether I may assume some calls won't throw, because then it's safe to combine the try/finally blocks. Applies here as well, would love to hear your view on this.

Sally7/Internal/NetworkStreamExtensions.cs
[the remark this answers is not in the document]
gfoidl: I'd remove that attribute.
* the JIT will inline this small method pretty sure
* makes the code cleaner

In general I apply _AggressiveInlining_ only when benchmarks + looking at codegen show it's beneficial. Otherwise leave the JIT alone with it's decisions.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: It's not very nice, but I think the only clean solution would be to perform additional reads when there's no pending request for a job ID that was received. That does impact a lot of the logic in the `ConcurrentRequestExecutor` though, so I'm not sure if that's what I want to do. This implementation is not very common of course, usually there's a separate receive thread for the incoming messages and pending requests are stored in a dictionary of sorts. That's definitely an option as well...

Sally7/Internal/CancellationTokenExtensions.cs
mycroes: You amaze me again with these details. When you're saying them they make absolute sense, but even when I see the line you comment on I'm still trying to figure what you'll say (before reading your comment).

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
gfoidl: Here this is basically the first statement in the `try-catch`, so that one can be collapsed --> do the dispose manually in the finally.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: I was actually quite happy with my solution, but I think yours still is better 😄

mycroes:
Thanks @gfoidl, some great comments, will make some changes.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
gfoidl: I'll think about an approach and share it with you (if I got some ideas).

Sally7/Internal/CancellationTokenExtensions.cs
gfoidl: Just look at the generated code, e.g. with ILSpy or any other de-compiler (like sharplab). Then you see the closures, etc.
With the suggestion they're gone and we are back to low allocations and high perf.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: One approach I have in mind:
- Add the jobID to `Request`
- Add state to `Request` [free, sent, received], not sure what states are ultimately desirable
- On receive:
  - Verify Requests[&lt;received job id> % &lt;number of concurrent requests>] has matching job ID
  - Interlocked.CompareExchange request state from sent to received
    - If this fails, drop the response
  - If job ID != request.jobId, perform another response read on the socket
  - Return request to pool
- On cancel:
  - Return request to pool
    - Increment job ID with &lt;number of concurrent requests>
    - Reset state
- Replace wait on `receiveSignal` with wait on `receiveSignal` or `request.Completed`

I also think the buffer could move to `Request`, because on request complete we process the buffer sync anyway. Last but not least I think the `sendSignal` isn't required, I think we're actually allowed to invoke `socket.Send...` concurrently, because we never send multiple frames.

Sally7/Internal/NetworkStreamExtensions.cs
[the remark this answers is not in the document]
mycroes: I generally only benchmark or look at codegen when stuff breaks 😆 But to that regard, I was actually recently looking into [ZeroMock](https://github.com/ZeroMock/ZeroMock) so I could mock the `TcpClient` / `Socket` in order to add unit tests without wrapping the socket implementation, [unfortunately that didn't work out](https://github.com/ZeroMock/ZeroMock/issues/15).

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: Changed this per recommendation. Considering all the safety measures in place for things that shouldn't happen (like checking if the memory is array based, which could theoretically pose an issue if someone would inject a different memory pool) it does feel like cutting corners, because an exception during `Dispose` would never release the `sendSignal`. I guess we can safely assume that never happens though, still somewhat bothers me 😉

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: Changed this per recommendation. I did figure another option, I could add `cancellationToken.Throw` as extension, add `[DoesNotReturn]` there and _maybe_ add the fallback exception as well (because someone could invoke that extension even when cancellation was not requested). Let me know what you think about that.

mycroes:
Made all the recommended changes, if you guys would be willing to do another review that would be awesome. I intend to release this shortly, the what-if fixes for the `ConcurrentRequestExecutor` can come at a later time.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
gfoidl: Hm, to me this sounds like a bit of ceremony for an easy solution that you have now (with initialized length). I wouldn't do this via extension method.

Side note: AFAIK `[DoesNotReturn]` is used only by C# for nullability static analysis (so it has no real effect in this case).

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
gfoidl: This approach sounds very clean 👍🏻 

As a more or less hacky solution one could catch and swallow the `ObjectDisposedException`, as it should only ever happen once canellation happened (which is asumed to be rare).

gfoidl (approved):
Just a minor thing, otherwise LGTM (the rest as you commented).

Sally7/Internal/SocketHelper.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Net.Sockets;
+
+namespace Sally7.Internal
+{
+    internal static class SocketHelper
+    {
+        public static void CloseSocketCallback(object? state)
+        {
+            var socket = state as Socket ?? throw new ArgumentException(
```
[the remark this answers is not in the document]
mycroes: I've never really used `Debug.Assert`, but I guess I totally agree with what you're saying. One issue is that there are nearly no unit tests in Sally7, that's something I want to change. Preferably without swapping out `Socket` for a wrapper...

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: I believe `[DoesNotReturn]` actually does more, see for instance S7Connection.cs `GetExecutorOrThrow()` method.

mycroes:
Thanks @gfoidl, again, I appreciate it a lot that you took the time to review this. Also @scamille thanks again for initial review! I hope you guys will stick around and provide feedback in the future as well.

gfoidl:
> stick around and provide feedback in the future as well.

Sure! Nice work is happening here 😃.

PR #38, opened 2023-06-22
Serialize Tpkt, Data and Header using Unsafe.WriteUnaligned

mycroes:


Sally7/Internal/WireFormatting.cs
```diff
@@ -0,0 +1,67 @@
...
+        return 3;
+    }
+
+    public static int WriteJobRequestHeader(ref byte destination, int paramLength, int dataLength)
+    {
+        WriteInt32(ref destination, JobRequestHeader1);
+        // Legacy, the
+        WriteInt16(ref destination.GetOffset(4), 1 << 8); // PDU ref
+        WriteInt16(ref destination.GetOffset(6), (short)paramLength);
+        WriteInt16(ref destination.GetOffset(8), (short)dataLength);
```
gfoidl, line 51: This code is nice and very readable.

Perf-wise we have 4 memory writes -- which may or may not be coalesced by cpu's store buffer.
So maybe combine these writes into a `long` + `short` (sure taking byte order into account), so we reduced that to 2 memory writes.

I'm not sure if a micro-benchmark will show any difference here. More likely a tool like Intel VTune can show differences in memory operations.

Further I believe that the code can be structed in a way that's still readable and understandable.
Maybe we should look into this in a follow-up PR.

Sally7/Internal/WireFormatting.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    public static int WriteData(ref byte destination)
+    {
+        // For performance we write an int instead of separate values, the final byte of the int will be overwritten.
+        WriteInt32(ref destination, Data);
+
+        return 3;
+    }
+
+    public static int WriteJobRequestHeader(ref byte destination, int paramLength, int dataLength)
```
gfoidl, line 45:
```suggestion
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    public static int WriteJobRequestHeader(ref byte destination, int paramLength, int dataLength)
```
My gut tells me that it's worth it here.
For the JIT this method seems probably too big to inline, but the actual code is quite little, so we should give the JIT a hint.

Sally7/Internal/TypeExtensions.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Runtime.CompilerServices;
+
+namespace Sally7.Internal;
+
+internal static class TypeExtensions
+{
+    public static ref byte GetOffset(this ref byte destination, int offset)
+    {
+        return ref Unsafe.Add(ref destination, offset);
```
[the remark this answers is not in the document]
mycroes: Thanks, I think you actually made a similar comment last week on AdsClient, I'll try to keep this in mind!

Sally7/Internal/WireFormatting.cs
mycroes: I'll add a benchmark. I also have a separate [repository for benchmarks](https://github.com/mycroes/Benchmarks), I'll see if I can put it into a usable form there as well. Either way I definitely agree, less writes is faster so it probably helps to combine it into a long + short write.

Sally7/Internal/WireFormatting.cs
mycroes: Do you know an easy way to verify if it will be inlined? I could at least benchmark the method call with and without `AggressiveInlining` to see if it makes a difference.

Sally7/Internal/TypeExtensions.cs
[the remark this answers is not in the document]
mycroes: Also, since offset will always be positive anyway, as will all the lengths returned from `WireFormatting` methods, I could just as well move to `uint` in all those places I guess.

mycroes:
Thanks again @gfoidl. Very nice that you jumped straight in, will keep the comments in mind for additional changes I intend to make as well.

Sally7/Internal/WireFormatting.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        WriteInt32(ref destination, JobRequestHeader1);
+        // Legacy, the
```
mycroes, line 48: Whoops, needs to be fixed.

Sally7/Internal/TypeExtensions.cs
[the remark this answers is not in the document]
gfoidl: Yep, `uint` works too.
As said in the AdsClient, it depends on where the cast to `uint` happens. Once here in this method, or when getting the `uint`-casted length of the span (which is `int`). I prefer having only one cast.
But you'll see when you make the change.

Sally7/Internal/WireFormatting.cs
gfoidl: Nice repo -- I'll dig into this next week 😃.

Sally7/Internal/WireFormatting.cs
gfoidl: In general I do either
* use https://sharplab.io/ and call the method from a dummy method to see what machine code gets emitted
* create a simple console app with a dummy method and set the proper env-variables (see https://github.com/dotnet/runtime/pull/74392 for more info)
* create a benchmark and use the `[DisassemblyDiagnoser]` to the see the machine code

Here with [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5il3YBLAHbYoATwDcNek1YAlDsIyD8MFgGEI+AA6CANnwDKfAG6CwMXNOoyAzE1IMAYhAg0A3jQY+mDkRgMbGIAFLBcDMDiGDAMACZWGGgMAQw6Ytj4ADIwwgDmGAAWyalx2BjYOflFAJTevl7Uvs1MAOwMAOqCsK5Q+OUq+SwdUIIxAFIQwPIwAI4ciQASMNgJUGEwEQm4SWkZ2bkFxfHllYe1Ns0AvvU+t34pykFiAPoAksL6IjBxGxFRMXiiRKT3SUEyVSOIMCZQqkNq90aLV8xHaXR60H6GEGeWGowmUxm8yWKzW70+31+4SBO2SYIh52OsLO1UKNUuvhu1C5NACfFE+jkSEcnW6MF6WJxnnuOlGpnKsUgwh2j0CABUdABrQIAXgYdAQdDsLzodBNZtNNnuAHpbQwADy4Dj4foSAB8NrtABFTgwdlAOGAMCxPdaHdanS6xOIPU1fLaw/bYK7NbhY80Eww1YVBLg/RgA0GUnmIJ9xAwHACrAx9KW8skuCJsIKq8WGBBTHwAO74mLCENxnyZ+3W5NiVPp3yywTywFKlWpH0VBh6pz2+0MUgoBh2+GFe7NQj64DkU/kc2mh0b8jCu1vBLKQSNvgH3xHugn00Wr9XhgADkkHcwwABS9ABVBhhGdYA+AYbBhDiBgAFEAHk1RoGU5QVBh50CVJJmmOYFh2ZZVj4cgV31BA7DXDctyAhhgKgCAMAgSBBTeL1Xx8I8QirGoAFkrFwbA8hgNVxB0dQCKJYjAnXBgb0Au0hNwESxIYDBJJgDDByAxNI1dGNQyzQpYlAiDwj4XJLAHDM7RHQzo0nIcHLHKAJxM7NcwYeV9AWNszNgBguGgGsYDybAwHLWA8FLARTNiR0YEVcIdQAIi0YQwA4KBYGUWTEiQhAYBy1ioHS603QYfAIDiJ9BGrIpYlrLsXz0zNW1YlIHxULgYqsHR4qsZJmoYDgdCkqBImiWJc2EAByQIoowDhm30csOFwH44MCMbarUZQ7PjBzRxgFM00wmdsNwv1CmgQJgLiDgZgiPUKIUgDdOaYgHGYYVUhGMZxK1DA/hmwFtkGcpBFLaFwtZOo9KRZE2k6fEYA+DAaPBqGmxUOGs1Bhgj0MRGOR8Hk9N+oVVXR4Gl2wcHWzx0QCeEJHmhR5E7V6NI+FCvp4MsBg2tFjG4OEOmRB2Ul2wibb6UBPziNGsyQqbFtZvlzT1dSLsDBbWIO27XtcmO5EgZiLGcepVmYcJxn2W+1HUQrCmGCpn6/pvOmrZgGSiJJMj1mpFnEnx2HhHh+kDlZeHmT3TmGm4+nreUW3NhpaH2eSQPiRI0lyOdvT7IYHJIuitWdNL3x/axm9cYjtmo5YABxGAMBQnhtrBlAamSD6Nz/dkGIshhwlT+vlEbu3m4d/sO67nvO5CJAB4YEJcHuqAMBqWOk495pp4wWes/t9n2877uuF7kIR+SLed73xPGRL1O3fIOgPa5Zp7gAbSEkUOqbxdD6BCEA+6cRQE6H0ChHQ7MBAAEE8h5FgGpGcmMKTCBEHkGoABde4NN/p+wxvnOSpEyQfC+MIH4zNtYXyjjHfYe4E6nCToiKeGMbakCbjsSOhNyHBzWOQEuqMGIVyiuIauXDgYNyQHwnOrcl43zvv3Qev4R7KRAuBCemxZHp1Pgoue/CW6livsvW+q916P23g9feLC35Hzrtwmexjz7z0violeYMH6bzsbvGor9yYu2RJ/b+9wvYoh9gDJ4J8z7/AYZ4phd0Hq+WbAsZOPhuYtAAHKdy7NATUKEoBrBMKMZsggABefA8RyLcYogR0d0n+RgGI127RcDVJgBALgT97E/1CQ8EhgNXHY14WHJJpiF7wxVm0zhtcfD5IwIUjyJSyl8EEJUmpUA6mGMzlsZJhM5ntLCZ07pvSQgBFOZ7GgVM+RQAFLTMgWZtLFT7F0+K0pqYxL0YkwE3irFgyKD5SZkMjnNNSL0oFWSGA5J+u0akYFlTYF4CwJBcQqQeOmbnTeHBrnQt7jcrk9zlD8mbM8pwyzVnFNKcYTZ2z2o5OIb7Yg254nuP+bERhhMAmBBOQs8RyKRJov9siypeRaFYsOTiphqdmhsDGFlTsu9alvFwFkMYGBDBIQQlsqWAB+IITYJBMVUGMTBAgZgqu2rq+q8FaFqRCCchgIAWmZMGbYH5zz2VjIORDblELZkZPmcjVOwrUXqDFQKQQkq6EmKUXDeVvhFUYGVXwGIuz1WauxDqvV8EGBGsVaIU1ox8AWs7FamANqYB2v1Y63AzqQ01Fde6tpnrSWZqebkZ0DBVLqXEtpVtVZvnNCESqPUBo6DkDQPcJBYBNSUSnaQWdel52akZkuw0dhV3NDAttKAm7J2GlaHcoAA==) (some copy & pasting needed) one can see the forcing the inline will help.

W/o the inline it's a tail-call (note the `jmp` as last instruction of method `Foo.Bar`, whilst w/ inline it's nice code in method `Foo.Bar_Inlined`.

Sally7/Internal/NetworkOrderSerializer.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Buffers.Binary;
+using System.Runtime.CompilerServices;
+
+namespace Sally7.Internal;
+
+internal static class NetworkOrderSerializer
+{
+    public static void WriteInt16(ref byte destination, short value)
```
gfoidl, line 9:
```suggestion
    public static void WriteInt16(ref byte destination, ushort value)
```
Found while inspecting the machine code (see other comment).
This was also sign-extended which isn't needed, as signed/unsigned does not make a difference in writing the bytes. So a little perf-boost here.

Either make the parameter unsigned, or cast to `ushort` in the two places used in this method.

gfoidl: Final codegen should look like [this](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5il3YBLAHbYoATwDcNek1YAlDsIyD8MFgGEI+AA6CANnwDKfAG6CwMXNOoyAzE1IMAYhAg0A3jQY+mDkRgMbGIAFLBcDMDiGDAMACZWGGgMAQw6Ytj4ADIwwgDmGAAWyalx2BjYOflFAJTevl7Uvs1MAOwMAOqCsK5Q+OUq+SwdUIIxAFIQwPIwAI4ciQASMNgJUGEwEQm4SWkZ2bkFxfHllYe1Ns0AvvU+t34pykFiAPoAksL6IjBxGxFRMXiiRKT3SUEyVSOIMCZQqkNq90aLV8xHaXR60H6GEGeWGowmUxm8yWKzW70+31+4SBO2SYIh52OsLO1UKNUuvhu1C5NACfFE+jkSEcnW6MF6WJxnnuOlGpnKsUgwh2j0CABUdABrQIAXgYdAQdDsLzodBNZtNNnuAHpbQwADy4Dj4foSAB8NrtABFTgwdlAOGAMCxPdaHdanS6xOIPU1fLaw/bYK7NbhY80Eww1YVBLg/RgA0GUnmIJ9xAwHACrAx9KW8skuCJsIKq8WGBBTHwAO74mLCENxnyZ+3W5NiVPp3yywTywFKlWpH0VBh6pz2+0MUgoBh2+GFe7NQj64DkU/kc2mh0b8jCu1vBLKQSNvgH3xHugn00Wr9XhgADkkHcwwABS9ABVBhhGdYA+AYbBhDiBgAFEAHk1RoGU5QVBh50CVJJmmOYFh2ZZVj4cgV31BA7DXDctyAhhgKgCAMAgSBBTeL1Xx8I8QirGoAFkrFwbA8hgNVxB0dQCKJYjAnXBgb0Au0hNwESxIYDBJJgDDByAxNI1dGNQyzQpYlAiDwj4XJLAHDM7RHQzo0nIcHLHKAJxM7NcwYeV9AWNszNgBguGgGsYDybAwHLWA8FLARTNiR0YEVcIdQAIi0YQwA4KBYGUWTEiQhAYBy1ioHS603QYfAIDiJ9BGrIpYlrLsXz0zNW1YlIHxULgYqsHR4qsZJmoYDgdCkqBImiWJc2EAByQIoowDhm30csOFwH44MCMbarUZQ7PjBzRxgFM00wmdsNwv1CmgQJgLiDgZgiPUKIUgDdOaYgHGYYVUhGMZxK1DA/hmwFtkGcpBFLaFwtZOo9KRZE2k6fEYDAj4MBo8GoabFQ4azUGGF4jgAhqQxEY5HweT036hVVdHgaXbBwdbfHREJ4QkeaFHkTtXo0j4UK+ngywGDayWMbg4QmZEHZSXbCJtvpQE/OI0azJCpsW1m5XNO11IuwMFtYg7bte1yY7kSBmIseUXHqU5mGidZ9lvtR1EKxphg6Z+v6byZu2YBkoiSTI9ZqQ5xICdh4R4fpA5WXh5k915hpuOZ+3sadzYaWh7nkjD4kSNJciPb0+yGBySLoq1nSq98EOHYwG88djrn45YABxGAMBQnhtrBlAamSD6Nz/dkGIshhwizlvsfb53O9d/s+4Hof+5CJAx4YEItvuqAMBqJP0995pF+UZf85d7ne/7weuGHkIp+SA/cCPk+08ZSus+98gdBfZcmaPcAA2kJIodU3i6H0CESB904gwJ0PoFCOhuYCAAIJ5DyLANSM4YAfC+MIEQeQagAF17gM3+sHDGJc5KkTJEQyk7N9Z33jonfYe5U6nHToiBeGNW55y2KvIuDB6ERzWOQSuqMGK1yiuIBuAjgatxviInYcdSwP03s/beo9x6/inspEC4E56bGUTna+SAO4aK7lojeT8X673fofB6p8uG/wvs3QRS9rEr1sWvbRjjt5v33q44+NQf7U09siABQD7j+xRIHAGTwr5tz8fnGOASxHhMCBrGAGcfD8xaAAOX7l2aAmoUJQDWCYUYzZBAAC8+B4hUb4mxhcOG+WbAsGRXt2i4CaTACAXAQifzccAmJDwaGAx8Y7UgrDIaiK6eTJ4+TCkMGKc0MpGAKkeWqbUvgggGnNKgK0yxOMFn+M6UTdZXifDe0Gc0kZIQKaTO5LpPkUABSMzIFmbSxU+yDPitKemySzH/H1g4reYMig+Wjmw5ZRNUgjN0SffhTcHntGpGBZU2BeAsEwXEKkt8kUJzCRTVFw8+m0xoHTL5PyaF/J2XsqpNTjBHJOe1Yp1Cg7EG3GktRENYjsKJrk7p/kCkYtkbikSBKW4CkEHkYQPwOmaITlnZobAxhZU7MfFpbxcBZDGBgQwSEELHLlgAfiCE2CQTFVBjAIQIGYertrmvqvBFVakQjrIYCACVvT3lUPBfy7OmNc5XMyYi7JKzUjrOlajWV+L1AKoacq1V1z1VoE1b4bVGBdV8BiGcw1xrsRmotfBBgNrtWiHtaMfATrOwupgG6mAHrLXetwL6npBT/WBoKe8+lyh+TNgYLkZ0DBVLqXEtpftVZQXNAkSqPUBo6DkBzXpTBYBNSUTXaQTdzRt2alZnuw0dhD2+DAttKAp7V2GlaHSoAA), so no more sign-extending moves -- they are slow and not needed.

Sally7/Internal/WireFormatting.cs
gfoidl: So, lunch done and I gave it a try:
```c#
    [MethodImpl(MethodImplOptions.AggressiveInlining)]
    public static int WriteJobRequestHeader(ref byte destination, int paramLength, int dataLength)
    {
        ulong tmp;

        if (BitConverter.IsLittleEndian)
        {
            tmp  = (ulong)BinaryPrimitives.ReverseEndianness(JobRequestHeader1);
            tmp |= (ulong)BinaryPrimitives.ReverseEndianness((short)(1 << 8)) << 32;
            tmp |= (ulong)(uint)BinaryPrimitives.ReverseEndianness((ushort)paramLength) << 48;
        }
        else
        {
            tmp  = JobRequestHeader1;
            tmp |= (ulong)(1 << 8) << 32;
            tmp |= (ulong)(uint)paramLength << 48;
        }

        Unsafe.WriteUnaligned(ref destination, tmp);

        WriteUInt16(ref destination.GetOffset(8), (ushort)dataLength);

        return 10;
    }
```
* code is not so pretty
* I'm not sure if the big-endian part is correct (this is where I'm always struggling)
* as this is the only place where this long+short is used, I don't think a helper is needed (similar to the `NetworkOrderSerializer`-class)

Maybe you have a better idea.
[Codegen-wise](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCEBXAMy5il3YBLAHbYoATwDcNek1YAlDsIyD8MFgGEI+AA6CANnwDKfAG6CwMXNNqNmLRctXqAksr4QdJqOcvWaNCIYfKL6ckhMpAwA6oKwAGLQ+NgYKsIA5jQA3jQMeQw6UIKmKTAMkMK4GAxBDAAqOgDW1QC8DHQIdADMAPp0dH0D/Ta5+QD0EwwAPLgc+MkSAHyjeRNjDAAiKdgMVVAcYBgsKwxr02Oz82Liy9T5p5NTsAuNuLf3Z3UAFoK4uxj7Q41P4QYT6cQMLoMYDiYJ/fSg9JoBhcETYMIw4LAhgQUx8ADuRVSMGExzu40eY2eYle73yhWKpXKoKqNWUm22DDaUSmUwYpBQDwYABkSekMF8TvdCO1gOR5eRBv1pnzyBFJi4ACYklSovhS/Iyuhy/pDU0qhgADkkQoAChsAKoMYRzYB8BjYYSahgAUQA8nUAuS8gySliKqzagApCDAeQwACOHCsGAAEjBsNqoOQue0EF0eXyBXaoBAMBBIGEXBsDXkZQAKTEwACUAFkrLhsOkYHVxDp1DG44nk6zeQw1TbJu3cJ3uwwMH2YEGPo9LgsbidPl8yvanbBeLBhJYySv1lMLnN13TVo9qVBaZvJt9fgwSvpk9jt7AUdAGIZ0tgYAQrAeAsiw9TbtMuAwGUYD7i0ABEWhHhwUCHhg8ZJimPoIDAYAcOWUAIWMiwMPgECaoIep/BKZQIvi+rBg86xNvOEA1NqThcMBVg6CyVjIrRDAcDo/ZQNCsJlL8wgAOTVIBGAcOi4LCdB3opPOkHkWoygnhSZ5UjALxvCcoZMhG1S4F80DVLamocPGXC5jmY7Wsu+TEFCzARLU0REj2TQYPW+4SVi2pVGiKigsitSGBkErNicORMfcxAAOwxP5DpuBgBbBTATnhWkKSCNF9SBQwDYcEEzZxeKXzNjY9wAL7uXknnhGy1R+YIwRbBg2D5U5rFFZFpXCIlTHJfcK4MIk4liVwSSepYDAMWt/kesIXVslUGbehATnQToYhMm+I6CZBqKhKFZSHZpUnsviBgYndeJQISvXBKStaZb1MDZcoeUhaNohRcIyL9dgjVtTN6WQk1+StdQJwANrthKFEuLo+j1hj1matjOj6H6OjgwIACC6TpLAM7FDAbj6CIIjpM2AC6Jwdd5O09cEg6YSOab7XwQ23QwoMlWVtQnVA2D4KK8VfDF7KatsCv1ZN9zTTNeS8wDOXAwV4spmNZX88OKbppmfDkDDKUUiKMAAUBl1LvbutZTlaqixL4MsAA4jAGB+jw0FBSgzbIi5fKWo1dqOgw+6/XrgMYN7IMm2D40B0HIdcGH9ZIJHDD1hwVk2c2Mty+rCWIzrKde0gPuZ5LpKB8HodB/WsfIqX5dQBgzaqwNNcNSM7tMBl5B0HXDDI2j+NYzjeNBwTRMk2T42U9TtO4PTjPMxk7Oc15ao8/55tYVUVtZj0FOiyNLfg8r1RV/LYoSi/4tqx/DVJb9HAEQZHnLoceOs8hURLmwXqyF3rBCgCwFwuBhRfUMD6L0ghPSa3AdrcB+QMC6DyG0UuQDWbQNEBIW0RR8C9XpgIeM71oLoMop6YQHZ6yX0FjfG2ds8H3AIToSqxDAGImbOQ64VDVC0LxPQmAjCYDMMwcINhM56z1n7oPes0crTNmbBaAss88ECKESXERR9S41XEZQ6h0irAODkXwJhGDWHsL7tZAeldTrv0VnoscKA3IT2RngmA+hoK/VwXw4xuZOGW2FtmQx4DjGEGEaQ5sWiLSx30aQBJOskkpNERY5QnjZbePqhafxOS56w3uA6So2BeAsBTqEQQ6Q2GambhFLOZUBF22Tp7ZQ6cja+2zu3POBce6mI0UPH+PiwHgPhtPWe88mJczPr5fpacm4hUfp01uyIy7uOqOdFs/8J4ADkg74mgI0P0UAszeEwUzAAXnwRpGzBmFSfuNZExzeHzIynvF5h11GHL+XkZZqVT4+XZA3IGpAH6SWNrs5+wlai/NOXgi5GArn3lufcvgjzBAvIQbC3K8KM7Iu+a+dEyYwVwwBUSmAwKapLJoMsoIIR0SdTIPURcOFvp7xZNkE+nVtmItGZ3IKEoXxirCl8qW7JDr5yDtgvIETUoZRCrUzsDSKaanaRS4qKLCmDyVWHOlyN2XuCgDdbmPKsU4puXc4wBL0SMqgMKlZUKmCClJR8sWwyyoHJstS98Jypq/W1fU9QTS3WtJgAaoZ8qIa/XuNAjAsC+DwMQcg1BCjnHbQAPwMCseISRNCVAyPsfIxRLjVG/IYCAUNtLWUoy9dy31GzDbDURYGiGqL2ToojRPKNDTY1M3jYmz5lLoqpvyOmzNA9XlIJQakNBBaGDFtLeW2xsia0FpUbgesDam2/NbVa+BN0SRzAYNOWcPZFyNtup6+4MTWRtA6HQcgaATgUzAI0XMn7SA/qYn+xoUNAOdC6CBmp0EoAQY/Z0NKbKaBAA) it's only two memory writes at the expense of more register chaining. So hard to tell (w/o measuring) what's faster. In doubt I'd keep the code as is now, because with the NetworkOrderSerializer it's so beautiful to read.

Sally7/Internal/TypeExtensions.cs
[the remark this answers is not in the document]
mycroes: Bummer, this is net6, 7 and 8 only and based on conversion to `UIntPtr` / `nuint`. I wasn't even targeting net6 yet. So now I'm wondering whether I want to go `uint` all around and cast it to `int` on < net6 or the other way around.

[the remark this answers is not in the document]
gfoidl: Ah, then something like `Unsafe.Add(ref src, (IntPtr)(uint)idx);` works when the language given mapping `UIntPtr <-> nuint` isn't available (and the implicit cast from `uint -> nuint`).
So I'd keep the parameters as `int` and do the casts in that method.

Sally7/S7ConnectionHelpers.cs
```diff
@@ -112,7 +112,7 @@ private static void BuildRequestItem(ref RequestItem requestItem, IDataItem data
             requestItem.VariableType = dataItem.VariableType;
         }
 
-        private static int BuildS7JobRequest(Span<byte> buffer, int parameterLength, int dataLength)
+        private static int BuildS7JobRequest(Span<byte> buffer, ushort parameterLength, ushort dataLength)
         {
             ref var start = ref MemoryMarshal.GetReference(buffer);
             var len = parameterLength + dataLength + 17; // Error omitted
```
gfoidl: `len` here will be `int`, so the parameters should remain `int`. See [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBLAHYYGAMQgQAFAFdcAC2jDsaBjPlRhwAJQMAvAD4G2BgGoGwANxce9BoOFjJdw8qdbdBo6Ys0AvkA==).

That's why -- from my experience -- it's quite often better to keep the parameters CLS compliant or use `uint` (as it's a native size that cpu registers can understand), and do the cast / type interpretion only where it's needed.
Squeezing out perf unfortunately comes with some traps too.

gfoidl (commented):
Left some comments. FYI: I'll be back on Monday -- wish you a nice weekend.

Sally7/S7ConnectionHelpers.cs
mycroes: Thanks for this one, makes sense but I didn't do any checking when changing to ushort...

Sally7/S7ConnectionHelpers.cs
```diff
@@ -74,8 +74,8 @@ public static int BuildWriteRequest(Span<byte> buffer, ReadOnlySpan<IDataItem> d
                 S7ProtocolException.ThrowDataItemCountExceedsParameterCount(dataItems.Length, parameters.Length);
             }
 
-            var fnParameterLength = dataItems.Length * 12 + 2;
-            var dataLength = 0;
+            var fnParameterLength = (ushort) (dataItems.Length * 12 + 2);
+            ushort dataLength = 0;
```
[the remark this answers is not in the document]
mycroes: Will change this back, I had my doubts around this part as well.

mycroes:
> Left some comments. FYI: I'll be back on Monday -- wish you a nice weekend.

A nice weekend to you as well! No need to hurry with reviewing, I'm not in a hurry to make these changes either. I'm aiming for GitHub activity every day, just so that I'll keep working on my projects and perhaps start some new projects too. I don't necessarily expect you to join my effort 😉

Sally7/Internal/WireFormatting.cs
mycroes: I guess this is hard to get your head around (still hurts my head as well), but reversing the endianess of the entire long suffices. It does make some sense, because for big endian all of the data is in the order that we want it on the network, so for little endian it's all in the incorrect order. That reduces the code complexity as well, and it's actually really fast (see benchmarks). Will definitely use long+short writes for this.

Sally7/Internal/WireFormatting.cs
```diff
@@ -0,0 +1,69 @@
...
+        WriteUInt32(ref destination, Data);
+
+        return 3;
+    }
+
+    public static uint WriteJobRequestHeader(ref byte destination, int paramLength, int dataLength)
+    {
+        var header = JobRequestHeader1 | (ushort) paramLength;
+
+        Unsafe.WriteUnaligned(ref destination,
```
scamille, line 51: Is there a reason you are not using WriteUint32 here? (assuming that is the correct type). Would encapsulate the operation bettter, in particular the conditional endian inversion.

Sally7/Internal/WireFormatting.cs
```diff
@@ -0,0 +1,69 @@
...
+            BitConverter.IsLittleEndian ? BinaryPrimitives.ReverseEndianness(header) : header);
+        WriteUInt16(ref destination.GetOffset(8), (ushort)dataLength);
+
+        return 10;
+    }
+
+    public static uint WriteUInt16(ref byte destination, ushort value)
+    {
+        NetworkOrderSerializer.WriteUInt16(ref destination, value);
+        return sizeof(short);
```
scamille, line 61: Is there something to access the compile time type of a variable in C# - which has no overhead?

Then you could just `return sizeof(decltype(value));` - making the code more generic with less chance of errors.

(I don't know why you currently don't use `sizeof(ushort)` - instead of adding the additional mental burden to know that sizeof for the same signed and unsigned type is the same)

scamille:
Interesting PR. While I am usually quite fond of doing optimizations, I am wondering if this really make a difference for such heavy I/O bound applications like you have with Sally7, even if you have tons of concurrent connections.

It also is sad that you have to go  through these extra steps and call `Unsafe` operations to squeeze out that extra performance :-) In particular since it seems to sacrifice some of the nice Span interface in favor of decoupled pointer/length arguments.

Sally7/Internal/WireFormatting.cs
mycroes: I need to have something to fix for the next day, don't I? 😜 It's actually `UInt64` / `ulong`, but yes, that's a definite improvement to make just like the other `WireFormatting` calls.

Sally7/Internal/WireFormatting.cs
mycroes: This started out as `short value`, thus return `sizeof(short)`. Forgot to clean these up as well. Unfortunately it's not possible to do `sizeof(value)`, but I bet there's quite some downsides to having that support in C#.

mycroes:
> Interesting PR. While I am usually quite fond of doing optimizations, I am wondering if this really make a difference for such heavy I/O bound applications like you have with Sally7, even if you have tons of concurrent connections.

That's a fair question. Performance wise there's quite some gain in the processing time, that doesn't change the I/O bounds. Still it could result in overall lower response times. Let's say it's better for the environment because we waste less CPU cycles. 💚 

> It also is sad that you have to go through these extra steps and call `Unsafe` operations to squeeze out that extra performance :-) In particular since it seems to sacrifice some of the nice Span interface in favor of decoupled pointer/length arguments.

I'm not sure if _Unsafe operations_ is the right definition. It's actually different from using the `unsafe` keyword, though I guess the effects are roughly the same in the sense that you could have buffer overflows / underflows.

And yes, one of the features of the library was the fact that everything was mapped to `struct`s and `enum`s. I actually still have the following sentence in the README:

> All protocols are mapped to `struct`s and `enum`s with the intent to create a project that is easy to comprehend and extend.

I'm going to divert from that approach. Basically because I'm dictating the direction of Sally7 and it's up to me to chose, but also because I value the performance gains. For me personally it's also unsellable that I'm writing more performant code in the next library ([AdsClient](https://github.com/VisconFactoryIntelligence/AdsClient)) while performance is **the** main feature of Sally7. However (and that's a BIG however), I don't want to turn this into a project that is unreadable, unmaintainable and in the end unusable. This should be _clean code_ and still have a relatively low barrier of entrance, it just won't be using structs for that in the long term.

I also invested a bit of time in unit tests, I want to expand on that. I'd like to add server side support as well at some point, not sure yet how I can do that while sharing as much of the code as possible. Maybe I will end up using `struct`s and code generation to get the best of both worlds. 😆 

Either way, thanks for your comments, again dearly appreciated! ❤️

scamille:
> That's a fair question. Performance wise there's quite some gain in the processing time, that doesn't change the I/O bounds. Still it could result in overall lower response times. Let's say it's better for the environment because we waste less CPU cycles. 💚

Processing time improvement can definitely be worthwhile - I just never was that big of a deal for the projects I worked with back in the days, since one side talked to a database and the other to a S7 PLC :-)

And yeah it is both interesting in how you can squeeze more performance out of .NET, and sad that you have to know these special tricks at the same time.

The AdsClient looks quite interesting. I worked on a small project using TwinCat at some point, and the library from Beckhoff that was already in place seemed to be okay. But that was with local communication only anyway.
Would you mind sharing (and maybe adding to the Readme) what the motivation was for creating your own library?

mycroes:
> > That's a fair question. Performance wise there's quite some gain in the processing time, that doesn't change the I/O bounds. Still it could result in overall lower response times. Let's say it's better for the environment because we waste less CPU cycles. 💚
> 
> Processing time improvement can definitely be worthwhile - it just never was that big of a deal for the projects I worked on back in the days, since one side talked to a database and the other to a S7 PLC :-)

Yeah I guess Viscon is by far the biggest user of Sally7 if you factor in the number of requests. Then again our abstraction layer still needs some serious work as well, that's not allocation free at all...

> And yeah it is both interesting in how you can squeeze more performance out of .NET, and sad that you have to know these special tricks at the same time.

I especially love all the details Günther is providing. I'm just using API's that are there, Günther is actually giving advice that squeezes every last bit of performance out of the CLR.

> The AdsClient looks quite interesting. I worked on a small project using TwinCat at some point, and the library from Beckhoff that was already in place seemed to be okay. But that was with local communication only anyway. Would you mind sharing (and maybe adding to the Readme) what the motivation was for creating your own library?

Let's call it ignorance 😄 I'm not familiar with Beckhoff PLC's, but we had another supplier at a customer that uses Beckhoff PLC's and we needed to do some communication. We asked their preferred form of communication and it was ADS. Searching for solutions I found the AdsClient library (I didn't write this from scratch). I think I did see NuGet packages for the Beckhoff library, but because I couldn't find any more information (should've just clicked the link on nuget.org though) I was a bit uncertain about using it.

So then I started by contacting the maintainer of the AdsClient library and got the GitHub repository transferred to me. The NuGet package was still linked to the original author though. He no longer had control due to account changes on nuget.org, but with a lot of communicating back and forth with NuGet admins I got that transferred too (not actually using it now though).

With all that out of the way I started doing some initial communication tests. That didn't go as planned at first, but soon enough we figured it was a similar issue to Siemens where both the IP and identifiers have to match. I could borrow a PLC though, so that allowed me to test locally and refactor the library into what it is now. There's still a lot of room for improvement there as well and just like Sally7 it's a bit short on unit tests, but it was fun to do and it works pretty well. Of course I still need to compare it against Beckhoff's library to see which one performs better and if it's the Beckhoff library I've got some work to do. Final thing I need to handle is multiple connections, which apparently is constrained to a single connection per IP, so I will be adding a proxy that solves that problem and is actually different from the router as provided by Beckhoff as well.

Long story short, I'm not sure I needed to write the library. But I did, and while doing that I even fixed a bug in the NodeJS ads-client package as well (which I also used for reference). And I got to know about `Unsafe.WriteUnaligned`, `Marshal.GetReference` and sneaky `Unsafe.As` usage. So a big win for me and a good reason to start some new work on Sally7!

gfoidl (approved):

PR #40, opened 2023-07-06
Data serialization

mycroes:


Sally7.Benchmarks/Serialization/SerializePrimitivesAsLargerPrimitive.cs
```diff
@@ -15,9 +15,9 @@ public void SerializeOneByOne()
     {
         ref var destination = ref MemoryMarshal.GetReference(buffer.AsSpan());
         uint offset = 0;
-        for (var i = 0; i < value.Length; i++)
+        foreach (var val in value)
```
gfoidl, line 18: Ah, the reason is that `value` is a field and not a local. So with the `foreach`-loop the C# compiler pulls the field into a local thus better code can be generated.

BTW: that's one of the reasons why I prefer the `_` prefix for fields, to easily distinguish between fields and locals. The in the previous code (with the `for`-loop) it would be obvious (for me) 😉.

mycroes: Yeah I need to start using `_` prefix. Years ago my manager decided we shouldn't use underscore prefixes, I've been regretting that decision for a while now. I stuck to that convention for my private projects as well, but I totally agree with what you're saying.

What I though was kinda funny was dat for the `...Combined` method the `foreach` didn't make any difference at all, the JIT is still generating the same instructions.

gfoidl: I think the JIT is quite sensitive here, or in other words it's luck when the JIT generates the same code here.
Some reasons for this are that the JIT can prove the field-access is safe, null-checks, etc. Best and most readable option here (array field) is the `foreach`-loop.

Sally7/ValueConversion/ToS7Conversions.cs
```diff
@@ -189,5 +180,46 @@ private static int ConvertFromString(string? value, int length, Span<byte> outpu
...
 #endif
         }
+
+        private static int CopyBuffer(ReadOnlySpan<byte> input, Span<byte> output)
+        {
+            ref var destination = ref MemoryMarshal.GetReference(output);
+            ref var source = ref MemoryMarshal.GetReference(input);
+
+            var offset = 0u;
+            while (offset + sizeof(ulong) <= input.Length)
```
gfoidl, line 190:
```suggestion
            while (offset <= input.Length - sizeof(ulong))
```
`offset` is incremented at the end of the loop, so `offset + sizeof(ulong)` needs to re-evaluated, whilst `input.Length - sizeof(ulong)` can be evaluated only once and kept in a register and compared to another register value (`offset`).

Sally7/ValueConversion/ToS7Conversions.cs
```diff
@@ -189,5 +180,46 @@ private static int ConvertFromString(string? value, int length, Span<byte> outpu
...
+            {
+                var value = Unsafe.ReadUnaligned<ulong>(ref source.GetOffset(offset));
+                NetworkOrderSerializer.WriteUInt64(ref destination.GetOffset(offset), value);
+
+                offset += sizeof(ulong);
+            }
+
+            if (offset + sizeof(uint) <= input.Length)
```
gfoidl, line 198: Here it doesn't really matter, but I'd write it as
```suggestion
            if (offset <= input.Length - sizeof(uint))
```
to be consistent with the loop above.

Same for the other checks below.

Sally7/ValueConversion/ToS7Conversions.cs
mycroes: Thanks again! I wasn't really happy with that condition to start with, this makes it feel a bit better at the least.

mycroes:
I hugely messed up on the 'serialize as larger primitive', that's not going to work. However I found https://github.com/dotnet/runtime/pull/76339, which gives me some new ideas to try out. As a coincidence (or not) @gfoidl is involved again 🤣

gfoidl:
That's pure coincidence 😉

Note: ATM I'm not actively thinking about the issue here, but please feel free to ping me if you'd like review, tipps, etc. (anything where you think I can help out).

mycroes:
Will do, for now I have to rework my benchmarks as well because of my mishap with trying to encode data using larger primitives, it makes me feel terribly stupid 🤦🏻‍♂️ . I need to look into `Vector128` and see what that offers for me, seems that might give me some performant solutions to everything serialization related in Sally7.

I think I should perhaps just follow everything @gfoidl on GitHub, maybe that leads to solutions for any problem I have 😉 On a (slightly) more serious note, I do enjoy seeing your name appearing in issues or PR's that interest me, that also makes me proud that you're spending time on reviewing Sally7 as well.

gfoidl:
Thanks ❤️ 

Well I use Sally7 for a business project...so I'm interested in what's going on here. Hooray for open source!
And the discussions in the PRs we had were very intersting and helpful. I think we both learned quite a lot and that should continue.

Sally7.Benchmarks/Serialization/SerializeArray.cs
```diff
@@ -0,0 +1,110 @@
...
+        [ParamsSource(nameof(ValueParameters))]
+        public ushort[] Value { get; set; } = Array.Empty<ushort>();
+
+        public IEnumerable<ushort[]> ValueParameters =>
+            Enumerable.Range(1, 12).Select(l => Enumerable.Range(1, l).Select(x => (ushort)x).ToArray());
+
+        [Benchmark(Baseline = true)]
+        public int SerializeOneByOne()
+        {
+            ref var destination = ref MemoryMarshal.GetReference(buffer.AsSpan());
```
gfoidl, line 26: Starting with .NET 5 you can also use
```suggestion
            ref byte destination = ref MemoryMarshal.GetArrayDataReference(buffer);
```

Sally7/ValueConversion/BufferHelper.cs
```diff
@@ -0,0 +1,109 @@
+﻿using Sally7.Internal;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Sally7.ValueConversion;
+
+internal class BufferHelper
+{
+    public static int CopyBytes(ReadOnlySpan<byte> input, Span<byte> output, int numberOfBytes)
```
gfoidl, line 10: For the product code why not just `input.CopyTo(output)`? This is a general purpose optimized version within .NET.
Is the specialized version of copy here as we know that only small sizes will be copied, etc.?


BTW: `numberOfBytes` should be given implicit via the spans (i.e. they should be sliced before calling this method).

Sally7.Benchmarks/Serialization/SerializeArray.cs
mycroes: Sounds like a good idea, at least for the benchmarks. I'm not sure about what frameworks to target, I intend to at least support the frameworks Microsoft supports, so I have to figure a nice way to support this across all targets. For the benchmarks I can perhaps live with just net5+ targeting, but I was already running into `Vector128` features that are available from net7+ only.

Sally7/ValueConversion/BufferHelper.cs
mycroes: > For the product code why not just `input.CopyTo(output)`? This is a general purpose optimized version within .NET. Is the specialized version of copy here as we know that only small sizes will be copied, etc.?

No I just messed up I guess. I started with one copy method and I had the idea `input.CopyTo(output)` is slow, because it doesn't do anything optimized. I once read about copying optimizations but I don't know what they applied to and I couldn't find that information anymore (and didn't spend much time searching, or checking the implementation in dotnet/runtime). However the perceived slowness might very well be related to the data I passed in. I should've benchmarked this as well, and if there are gains in small optimizations I can still do a different implementation than the one in the BCL, but preferably I just use what's already there. From net8 onwards there's also optimized versions of `BinaryPrimitives.ReverseEndianness` for `span` (the dotnet PR referenced above that you also reviewed), I want to use those as well when available.

> BTW: `numberOfBytes` should be given implicit via the spans (i.e. they should be sliced before calling this method).

Another mishap during refactoring, so yes, you're absolutely right (and they should also be properly sliced in current Sally7 code, I guess only benchmarks are an exemption right now).

gfoidl: `Span<T>.CopyTo` in .NET needs to be general purpose, i.e. can handle all given length as optimized as possible.
This -- of course -- induces some length check and a bigger non inlineable method.

If user code knows that input is always below a specific size or has constant size, then it's quite easy to write a CopyTo that is faster -- but also specialized for that use case.

I'd go with the builtin method, as it's well tested and safe. Then only switch to a custom version if benchmarks show it's worth it.
(Just writing the last sentence for completeness, as I'm aware we have the same thinking here)

mycroes: I added a benchmark comparing my implementation against `Span.CopyTo`, the latter was faster for arrays with power of two lengths, so I just favored `Span.CopyTo`. There's room for improvement in mine, but I don't think that would actually turn the tables.

mycroes:
@gfoidl Care to take a final look at this? I think it can be merged as-is.

Sally7/ValueConversion/BufferHelper.cs
gfoidl: > for arrays with power of two lengths

Yep, it's due to copying via `short` (2), `int` (4), `long` (8), `Vector128<byte>` (16), `Vector256<byte>` (32).

But thanks for measuring.

mycroes: > > for arrays with power of two lengths
> 
> Yep, it's due to copying via `short` (2), `int` (4), `long` (8), `Vector128<byte>` (16), `Vector256<byte>` (32).

Which is exactly what I was doing, but only up to long 😉

Sally7/ValueConversion/BufferHelper.cs
```diff
@@ -0,0 +1,69 @@
...
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static int CopyAndAlign64Bit(ReadOnlySpan<byte> input, Span<byte> output, int numberOfItems)
+    {
+        ref var destination = ref MemoryMarshal.GetReference(output);
+        ref var source = ref MemoryMarshal.GetReference(input);
+
+        var limit = (uint) numberOfItems * sizeof(ulong);
+
+        var offset = 0u;
+        while (offset < limit)
```
gfoidl, line 19: This assumes that after the loop no remaining elements in the input-span are present. This is OK for the use case here (?), but would you mind adding an assert for this? E.g.
```c#
Debug.Assert(offset == input.Length, "input span's length is not a multiple of sizeof(ulong)");
```
Then the code is self documenting.

gfoidl (approved):
LGTM -- just one suggestion.

I had only a look at the product code. The benchmarks were just a quick look.

Sally7/ValueConversion/BufferHelper.cs
mycroes: I'm afraid I need your advice on this. I cheated the input of these methods by doing `Unsafe.As<long[], byte[]>(ref value)`. As a result, for an array of 8 `long` values, I get a `byte[]` with it's `Length` set to... 8. Then in `CopyAndAlign64Bit` I take the pointer to that buffer and just copy the actual number of bytes needed.

So the question is, did I mess up? I could use `MemoryMarshal.Cast<long, byte>(value)`, but if you look at the [benchmarks](https://github.com/mycroes/Benchmarks/commit/ebe032a118c1c738ea02094ab679fb69c1930ccd), there's quite some difference. But using `Unsafe.As<,>()`  means I'm messing with the length, and as such I lose the validation. As a workaround, I could pass in the original array and apply the conversions in the method, but still I won't have `input.Length` that matches the actual length, so the only approach I can take is to calculate based on `numberOfItems`, which is the same calculation as `limit` is ergo I'm validating if I can do the same calculation twice...

Let me know what you think. I think I currently favor the existing approach that relies on me writing the code correctly, but for sure I can't do that 100% of the time.

Sally7/ValueConversion/BufferHelper.cs
```diff
@@ -0,0 +1,69 @@
+﻿using Sally7.Internal;
+using System;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Sally7.ValueConversion;
+
+internal class BufferHelper
```
gfoidl, line 8:
```suggestion
internal static class BufferHelper
```

Sally7/ValueConversion/BufferHelper.cs
gfoidl: You're right, and maybe I was a bit too conservative here -- all in all it's an internal type and we are under full controll what's passed in. So (now) I think just a little comment like
```c#
// The data passed in is guaranteed to be powers of two long, so no remaining elements will exists.
```
or similar.

In general and where possible I prefer the `Debug.Assert` as it docuemts the code, and that assert can be checked in CI by a debug build too. But as you said, it's not easy to pass in the required info w/o writing more code than needed.

PR #48, opened 2023-11-30
fix: Prevent exception on return of job ID to disposed pool

mycroes:
Currently when the connection is disposed and requests are still in the queue they will except because the job ID can no longer be returned to the pool. Making the exception on `ReturnJobId` conditional on a new `_disposed` flag prevents the exception from being thrown, but it does permit 'successful' returns to a disposed pool.

@gfoidl please let me know what you think. This is hitting us in production now, so I'll probably merge this quickly.

### DIFF
--- a/Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
+++ b/Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
@@ -192,7 +192,7 @@ namespace Sally7.RequestExecutor
         {
             private readonly Channel<int> _jobIdPool;
             private readonly Request[] _requests;
-            private volatile bool _disposed;
+            private bool _disposed;
 
             public JobPool(int maxNumberOfConcurrentRequests)
             {
@@ -212,7 +212,7 @@ namespace Sally7.RequestExecutor
 
             public void Dispose()
             {
-                _disposed = true;
+                Volatile.Write(ref _disposed, true);
                 _jobIdPool.Writer.Complete();
             }
 
@@ -220,7 +220,7 @@ namespace Sally7.RequestExecutor
 
             public void ReturnJobId(int jobId)
             {
-                if (!_jobIdPool.Writer.TryWrite(jobId) && !_disposed)
+                if (!_jobIdPool.Writer.TryWrite(jobId) && !Volatile.Read(ref _disposed))
                 {
                     Sally7Exception.ThrowFailedToReturnJobIDToPool(jobId);
                 }
--- a/Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
+++ b/Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
@@ -188,55 +188,6 @@ namespace Sally7.RequestExecutor
             }
         }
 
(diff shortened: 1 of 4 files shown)


mycroes:
@gfoidl No need to hurry, hotfixed the previous release, because there's some unreleased changes on main anyway.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
```diff
@@ -192,6 +192,7 @@ private class JobPool : IDisposable
         {
             private readonly Channel<int> _jobIdPool;
             private readonly Request[] _requests;
+            private volatile bool _disposed;
```
gfoidl, line 195: Just a minor suggestion based on my preference and experience.

Instead of using the `volatile` modifier I prefer using explicit `Volatile.Read(ref)` / `Volatile.Write(ref, value)` as it's more obvious when reading the code. It's about visibility of the data, thus an important aspect (to prevent / hunt some bugs).

gfoidl (approved):
Nice and simple solution.

Makes sense!

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
mycroes: I agree, marking the field `volatile` doesn't solve the visibility of the issue.

mycroes:
Thanks @gfoidl, appreciate your opinion a lot. I intend to add some unit tests for this behavior as well, so I'm not merging straight away.

Sally7/RequestExecutor/ConcurrentRequestExecutor.cs
```diff
@@ -209,13 +210,17 @@ public JobPool(int maxNumberOfConcurrentRequests)
...
+                Volatile.Write(ref _disposed, true);
+                _jobIdPool.Writer.Complete();
+            }
 
             public ValueTask<int> RentJobIdAsync(CancellationToken cancellationToken) => _jobIdPool.Reader.ReadAsync(cancellationToken);
 
             public void ReturnJobId(int jobId)
             {
-                if (!_jobIdPool.Writer.TryWrite(jobId))
+                if (!_jobIdPool.Writer.TryWrite(jobId) && !Volatile.Read(ref _disposed))
```
scamille, line 223: Is there a reason for the ordering? Wouldn't you want the `!Volatile.Read(ref _disposed)` checked first, and the TryWrite only executed if it isn't yet disposed, taking care of short-circuiting `&&`.

Or is there a reason for always executing the writing to the jobIdPool, independent of it being disposed or not?

gfoidl: TryWrite is called more often than Dispose so I think that order is fine, meaning that the common path comes first.
If TryWrite fails, then it's checked if disposed or not.

But TBH I don't think that any order here actually matters. The arguing of @scamille is also fine.
Both yield the same result, and perf-wise there's no difference (at least with TCP connection to PLC in mind). And I have no problem if the order is flipped.

PR #49, opened 2024-02-12
Hotfix conversion of float/double

mycroes:
Fix conversion of single `float`/`double` values to their PLC format. The assumption that the runtime would treat `Action<float>` similarly to `Action<int>` is apparently invalid, as can also be witnessed from [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxBAzmAPgAQEwEYCwAUJgMwAEOZAwmQN5FmMXmYAsZAsgBQCUdDTAL4DGI5hXYB1AE5QEAUwCSMBF1gIyANwgAbAK7y+9Qk1MVcATi7b9hgNxjhJpmNISyMufIBiOgPYQqgBm/oFaugZGYqaYltYR9o5EgkA==).

### DIFF
--- a/Sally7/ValueConversion/ConverterFactory.cs
+++ b/Sally7/ValueConversion/ConverterFactory.cs
@@ -3,7 +3,7 @@ using System.Runtime.CompilerServices;
 
 namespace Sally7.ValueConversion
 {
-    public delegate int ConvertToS7<TValue>(TValue? value, int length, Span<byte> output);
+    public delegate int ConvertToS7<TValue>(in TValue? value, int length, Span<byte> output);
 
     public delegate void ConvertFromS7<TValue>(ref TValue? value, ReadOnlySpan<byte> input, int length);
 
--- a/Sally7/ValueConversion/ToS7Conversions.cs
+++ b/Sally7/ValueConversion/ToS7Conversions.cs
@@ -61,7 +61,7 @@ namespace Sally7.ValueConversion
         {
             if (value == null) throw new ArgumentNullException(nameof(value), "Value can't be null.");
 
-            return BufferHelper.CopyAndAlign64Bit(Unsafe.As<long[], byte[]>(ref value), output, length);
+            return BufferHelper.CopyAndAlign64Bit(Unsafe.As<long[], byte[]>(ref Unsafe.AsRef(value)), output, length);
         }
 
         private static int ConvertFromInt(in int value, int length, Span<byte> output)
@@ -75,7 +75,7 @@ namespace Sally7.ValueConversion
         {
             if (value == null) throw new ArgumentNullException(nameof(value), "Value can't be null.");
 
-            return BufferHelper.CopyAndAlign32Bit(Unsafe.As<int[], byte[]>(ref value), output, length);
+            return BufferHelper.CopyAndAlign32Bit(Unsafe.As<int[], byte[]>(ref Unsafe.AsRef(value)), output, length);
         }
 
         private static int ConvertFromShort(in short value, int length, Span<byte> output)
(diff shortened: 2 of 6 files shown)


Sally7/ValueConversion/ConverterFactory.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Sally7.ValueConversion
 {
-    public delegate int ConvertToS7<TValue>(TValue? value, int length, Span<byte> output);
+    public delegate int ConvertToS7<TValue>(in TValue? value, int length, Span<byte> output);
```
gfoidl, line 6: Is the `in` actually needed?
For small `TValue` (roughly 2x CPU register-size) this is actually a de-optimization, as the value has to be passed via stack rather than cpu's registers (as w/o `in`).

For value types the generic constraint `: struct` would guarantee the readonly-ness here too (copy by value).
If arrays need to be passed in, then I'd use `ReadOnlySpan` instead.

Sally7/ValueConversion/ToS7Conversions.cs
```diff
@@ -48,14 +48,14 @@ public static Delegate GetConverter<TValue>()
             throw new NotImplementedException();
         }
 
-        private static int ConvertFromLong(long value, int length, Span<byte> output)
+        private static int ConvertFromLong(in long value, int length, Span<byte> output)
```
gfoidl, line 51: Same (and on other places).

gfoidl (commented):
> The assumption that the runtime would treat Action<float> similarly to Action<int> is apparently invalid

Because for value types specialized code for the generics is created, with each optimized. Thus there may be differences.

Also on sharplab you can choose between x64, x86 and the old .NET Fx.

Sally7/ValueConversion/ConverterFactory.cs
mycroes: Depends on what you'd call needed. I could add separate `float`/`double` handling (under the assumption that'll work as expected in that case), but apparently there's a difference in how the framework treats `float`s vs `int`s, while I would expect that it would just pass 4 bytes around in the same way. The same also applies for `double` vs `long`.

So maybe it's a good idea to split `float` / `double` regardless since it removes the downsides, but it was definitely broken in the previous state.

gfoidl: For .NET 6+ I'd throw away the delegate and use an approach that is based on generic math (i.e. static abstract interface members).

* no problems with endianess (I believe that's the root issue here)
* superb perf (delegates aren't as optimized as struct's generic specialization)
* easy to extend

When we have some time with this I can look at the weekend into that approach.

mycroes: Your help is always greatly appreciated! I think main problem here is that the conversion also supports strings and arrays, so I'm not sure we can handle all in a uniform fashion. But perhaps it's a good idea to differentiate anyway. I must admit I haven't applied static abstract interface members anywhere yet, even though I was hugely looking forward to its implementation before .NET 6. Guess I've been too busy lately 😉

gfoidl: OK, when the test pass here let's go ahead and merge this.
I'll create an issue that reminds me on working with the static abstract interfaces for this. For strings and arrays I don't see a problem -- when there's a PR for this I hope you see why there shouldn't be a problem.

mycroes: I actually already released this as 0.14.4. The hotfix branch doesn't cleanly merge into main though, because I did some work on main that's not yet released. Like I said, been a bit too busy, will try to prepare a release of main sooner rather than later.

PR #53, opened 2024-03-12
Dispose requests on JobPool Dispose

### LINKED ISSUE #52: Stuck requests when receiveSignal is released after Dispose()
Consider following stack trace:
```
Sally7.Sally7Exception: Couldn't signal receive done.
   at Sally7.Sally7Exception.ThrowFailedToSignalReceiveDone()
   at Sally7.RequestExecutor.ConcurrentRequestExecutor.PerformRequest(ReadOnlyMemory`1 request, Memory`1 response, CancellationToken cancellationToken)
   at Sally7.S7Connection.ReadAsync(IDataItem[] dataItems, CancellationToken cancellationToken)
   at VLC.PLC.Sally7.Sally7PlcConnection.ReadIntermediateAsync(IEnumerable`1 readContainers)
   at VLC.PLC.PlcConnection.ReadAsync(IEnumerable`1 readContainers, Boolean triggerChanges)
   at VLC.PLC.Polling.PlcPoller.TimerCallback(Object state)
```

The following must have happened:
1. Request has entered `PerformRequest`
2. Request was sent
3. Response was received
4. `ConcurrentRequestExecutor` is disposed
5. `receiveSignal.TryRelease()` returns `false`
6. `Sally7Exception.ThrowFailedToSignalReceiveDone()` is invoked

If there was only one request, this results in the caller receiving the exception from point 6.

Now assume there was already a caller waiting:
1. Request A was sent
2. Request B was sent
3. Caller A receives response B
4. `rec.Complete` is called for `Request` B
5. Caller A awaits `Request` A (no cancellation)
6. Caller B receives response A
7. Caller B exceptions due to `receiveSignal.Dispose()`
8. Caller B receives the exception
9. `Request` A is never completed
10. Caller A is stuck

Possible solution:
1. On `JobPool.Dispose()`
2. `Dispose()` all `Request`s
   a. `Request.Dispose() { _disposed = true; _continuation.Invoke(); }`
   b. `Request.GetResult() { ObjectDisposedException.ThrowIf(_disposed, this); return ... }`
3. Add `ObjectDisposedException.ThrowIf` to `ConcurrentRequestExecutor` throw paths?

Disadvantages?

mycroes:
Fixes #52

### DIFF
--- a/Sally7/Internal/DisposableHelper.cs
+++ b/Sally7/Internal/DisposableHelper.cs
@@ -1,0 +1,13 @@
+using System;
+
+namespace Sally7.Internal;
+
+internal static class DisposableHelper
+{
+    public static void ThrowIf(bool condition, object instance)
+    {
+        void Throw() => throw new ObjectDisposedException(instance.ToString());
+
+        if (condition) Throw();
+    }
+}
--- a/Sally7/Internal/DisposableHelper.cs
+++ b/Sally7/Internal/DisposableHelper.cs
@@ -6,7 +6,7 @@ internal static class DisposableHelper
 {
     public static void ThrowIf(bool condition, object instance)
     {
-        void Throw() => throw new ObjectDisposedException(instance.ToString());
+        void Throw() => throw new ObjectDisposedException(instance.GetType().FullName);
 
         if (condition) Throw();
     }
--- a/Sally7/Internal/DisposableHelper.cs
+++ b/Sally7/Internal/DisposableHelper.cs
@@ -6,8 +6,12 @@ internal static class DisposableHelper
 {
     public static void ThrowIf(bool condition, object instance)
     {
+#if NET7_OR_GREATER
+        ObjectDisposedException.ThrowIf(condition, instance);
+#else
         void Throw() => throw new ObjectDisposedException(instance.GetType().FullName);
 
         if (condition) Throw();
+#endif
     }
 }
--- a/Sally7/RequestExecutor/Signal.cs
+++ b/Sally7/RequestExecutor/Signal.cs
@@ -1,0 +1,26 @@
+﻿using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Sally7.RequestExecutor
+{
+    [DebuggerNonUserCode]
+    [DebuggerDisplay(nameof(NeedToWait) + ": {" + nameof(NeedToWait) + ",nq}")]
+    internal class Signal : IDisposable
+    {
+        private readonly Channel<int> channel = Channel.CreateBounded<int>(1);
+
+        public void Dispose() => channel.Writer.Complete();
+
+        public bool TryInit() => channel.Writer.TryWrite(0);
+
+        public ValueTask<int> WaitAsync(CancellationToken cancellationToken) =>
+            channel.Reader.ReadAsync(cancellationToken);
+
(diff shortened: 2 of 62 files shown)


mycroes:
Please ignore the S7 conversion bits of this PR. It's currently deviating from main because #49 is also not yet merged. I would prefer to merge these two and then perhaps add the generic math / static abstract on top.

mycroes:
I'm releasing this as 0.14.5 BTW.

mycroes:
After some struggles this has been released as 0.14.6 instead.

Sally7/Internal/DisposableHelper.cs
```diff
@@ -0,0 +1,13 @@
+using System;
+
+namespace Sally7.Internal;
+
+internal static class DisposableHelper
+{
+    public static void ThrowIf(bool condition, object instance)
```
gfoidl, line 7: Starting with .NET 7 theres [ObjectDisposedException.ThrowIf](https://learn.microsoft.com/en-us/dotnet/api/system.objectdisposedexception.throwif?view=net-8.0).

Should we `#if`def it here so the built-in method can be used?

Sally7/Internal/DisposableHelper.cs
```diff
@@ -0,0 +1,13 @@
...
+    {
+        void Throw() => throw new ObjectDisposedException(instance.ToString());
```
gfoidl, line 9:
```suggestion
        void Throw() => throw new ObjectDisposedException(instance.GetType().FullName);
```
?

Sally7/RequestExecutor/Signal.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void Dispose() => channel.Writer.Complete();
+
+        public bool TryInit() => channel.Writer.TryWrite(0);
+
+        public ValueTask<int> WaitAsync(CancellationToken cancellationToken) =>
+            channel.Reader.ReadAsync(cancellationToken);
+
+        public bool TryRelease() => channel.Writer.TryWrite(0);
+
+        private bool NeedToWait => channel.Reader.Count == 0;
```
gfoidl, line 24: This method can't be used (it's `private` and the type isn't `partial`), so remove it?

Sally7/Internal/DisposableHelper.cs
mycroes: Not sure how we should go about declaring the static `ThrowIf` on net < 7? I took `ObjectDisposedException.ThrowIf` as example of course, just not sure how to create the compatibility here.

gfoidl: Via a
```c#
private void EnsureNotDisposed()
{
#if NET7_OR_GREATER
    ObjectDisposedException.ThrowIf(_stream is null, this);
#else
    // Custom method
#endif
}
```
Or make that method `public` in the helper type.

Sally7/Internal/DisposableHelper.cs
gfoidl: I looked at BCL 😉.

Sally7/RequestExecutor/Signal.cs
gfoidl: Ups, didn't see at that while reviewing. Thanks for the hint.
(and yes I added it)

Sally7/Internal/DisposableHelper.cs
mycroes: Could have guessed 😄

gfoidl (approved):

PR #65, opened 2026-04-17
Read write errors

mycroes:
Add detailed read/write error information and introduce an optimized path that no longer throws on data item errors.

### DIFF
--- a/Sally7/S7Connection.cs
+++ b/Sally7/S7Connection.cs
@@ -2,7 +2,6 @@ using System;
 using System.Buffers;
 using System.Diagnostics.CodeAnalysis;
 using System.Net.Sockets;
-using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Sally7.Infrastructure;
@@ -357,7 +356,7 @@ namespace Sally7
 
             Throw();
             [DoesNotReturn]
-            static void Throw() => throw new ArgumentException("Results array length needs to be at least the same as the data items length.", nameof(resultsLength));
+            static void Throw() => throw new ArgumentException("Results array length needs to be at least the same as the data items length.", "results");
         }
 
         private IRequestExecutor GetExecutorOrThrow()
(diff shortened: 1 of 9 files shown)


mycroes:
@copilot apply changes based on the comments in [this thread](https://github.com/mycroes/Sally7/pull/65#pullrequestreview-4127711067)

Sally7/S7Connection.cs
```diff
@@ -208,9 +209,51 @@ await stream
...
+            try
+            {
+                await ReadAsync(dataItems, results, cancellationToken).ConfigureAwait(false);
+
+                ReadWriteErrorHelpers.ThrowIfHasErrors("Read", dataItems, results.AsSpan(0, dataItems.Length));
+            }
+            finally
+            {
+                ArrayPool<ReadWriteErrorCode>.Shared.Return(results);
+            }
```
gfoidl, line 239: `try-finally` for the array pool usage isn't needed, actually it's rather an anti-pattern.
Array pool can handle dropped arrays quite well.

Sally7/S7Connection.cs
```diff
@@ -238,9 +279,51 @@ public async Task ReadAsync(IDataItem[] dataItems, CancellationToken cancellatio
...
+            try
+            {
+                await WriteAsync(dataItems, results, cancellationToken).ConfigureAwait(false);
+
+                ReadWriteErrorHelpers.ThrowIfHasErrors("Write", dataItems, results.AsSpan(0, dataItems.Length));
+            }
+            finally
+            {
+                ArrayPool<ReadWriteErrorCode>.Shared.Return(results);
+            }
```
gfoidl, line 309: Ditto.

Sally7/S7Connection.cs
mycroes: Thanks a lot @gfoidl! I was doubting what to do, because I myself might end up reading invalid addresses at high rate as well. Then again, I'll be using the overload accepting `Memory` to handle this, so I think it's fair to assume that anyone worrying about only consuming from `ArrayPool` and never returning will use the right overload as well.

Currently on a short holiday, but I will fix this when I'm back.

gfoidl: Enjoy your holiday!

